Issue #85, opened 2014-12-11
Scripts assume /bin/sh is POSIX compliant


lildude:
All the scripts in backup-utils assume the host on which they're being run has a bash-compatible `/bin/sh`.  This isn't the case on Solaris 10 and earlier where `/bin/sh` is the original Bourne shell which is not 100% compatible with bash.

To ensure better cross-platform compatibility, the scripts should explicitly use `#!/bin/bash`

snh:
Fixed by #177

PR #87, opened 2014-12-16
Add test to check for bashisms.

mikemcquaid:
As discussed here: https://github.com/github/backup-utils/issues/85#issuecomment-66679219

CC @rtomayko @lildude

### DIFF
--- a/test/test-bashisms.sh
+++ b/test/test-bashisms.sh
@@ -1,0 +1,12 @@
+#!/bin/sh
+# bashisms tests
+
+# Bring in testlib
+. $(dirname "$0")/testlib.sh
+
+begin_test "ghe-* bashisms"
+(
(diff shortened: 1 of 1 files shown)


mikemcquaid:
Repinging @rtomayko and @lildude for any thoughts. Happy with just a :+1: or :-1: on here.

lildude:
:+1: This won't pick up on the use of bash in [ghe-rsync](https://github.com/github/backup-utils/blob/master/share/github-backup-utils/ghe-rsync) but it will pick up any other unintentional bashisms in /bin/sh scripts.

mikemcquaid:
@rtomayko Let me know if there's any problems here. As it's just adding another (passing) test I might just merge this in a few days if that's alright.

PR #90, opened 2015-01-14
report mysqldump failures

buckelij:
`ghe-export-mysql` errors were not correctly reported because the gzip in the pipeline succeeds even when mysqldump does not.

This fixes mysql backup error reporting by switching from sh to bash and setting '-o pipefail'.


rtomayko:
Made a quick change to avoid `grep -q`. See https://github.com/github/backup-utils/commit/2acbdfd3c3022f98a9e85b4eec4f4125cb62ca1f for details. Looks great otherwise. Thanks!

PR #91, opened 2015-01-14
Use the new ghe-negotiate-version script

rtomayko:
Makes use of the new `ghe-negotiate-version` script to ship in GHE v2.1. The idea is to push more of the version checking logic to the appliance. This will eventually let us get rid backup-utils logic that assumes hard-coded version file paths and gives the appliance a chance to require a minimum backup-utils version in a controlled way.

Original patch by @mikemcquaid in our (now defunct) internal backup-utils repository.


mikemcquaid:
Looks good to me :+1:

PR #94, opened 2015-01-19
Allow ghe-host-check to be run from within the bin directory

lildude:
At the moment `ghe-host-check` assumes it'll always be run via `bin/ghe-host-check` from within the `backup-utils` directory or via `ghe-host-check` with `backup-utils/bin` in the user's path.

This fails with:

```
$ ./ghe-host-check
./ghe-host-check: line 54: /path/to/backup-utils/ghe-host-check: No such file or directory
$
```

... on OS X, which is useful, or uselessly:

```
$ ./ghe-host-check
./ghe-host-check: 54: exec: ./ghe-host-check: not found
$
```

... on Ubuntu 14.04 (reproduced on AWS with ami-0307d674).

This PR corrects that and allows for running `ghe-host-check` from within the `backup-utils/bin` directory without affecting the other methods of calling it.

### DIFF
--- a/bin/ghe-host-check
+++ b/bin/ghe-host-check
@@ -38,7 +38,7 @@ if [ $rc -ne 0 ]; then
     case $rc in
         255)
             if echo "$output" | grep -i "port 22: connection refused" >/dev/null; then
-                exec "$0" "$hostname:122"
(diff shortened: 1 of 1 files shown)


rtomayko:
Looks good to me. Thanks!

PR #96, opened 2015-01-28
Document the -c option

lildude:
Several people have been a little surprised to find their configuration isn't restored to the destination host as they know nothing about the `-c` option and I don't think many people query `ghe-restore` to find out the available options.

This PR documents this option.

### DIFF
--- a/README.md
+++ b/README.md
@@ -153,6 +153,8 @@ The `ghe-backup` and `ghe-restore` commands also have a verbose output mode
 (`-v`) that lists files as they're being transferred. It's often useful to
 enable when output is logged to a file.
 
+The settings and license are not restored by default to prevent overwriting different configuration on the restore host. Restore appliance settings and license in addition to datastores by passing the `-c` argument to `ghe-restore`
+
 ### Scheduling backups
 
 Regular backups should be scheduled using `cron(8)` or similar command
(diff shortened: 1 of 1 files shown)


rtomayko:
Yeah, the behavior here is a little weird. Settings are restored automatically if the instance has never been configured (migrate flow) but must be specified explicitly if restoring over an already configured instance. We'll need to incorporate that bit of nuance here.

tedivm:
This would have been good to know. Any reason this little tidbit hasn't been merged in yet?

rtomayko:
@tedvim Nope, just fell off my radar. I want to clarify the bit about new vs. already configured instances and then I'll get this merged down.

Xeago:
While we're at it, I'd like to make a feature request for overriding a single parameter when restoring: the hostname. The hostname is used to identify the instance within collectd - restoring a the configuration with `github.example.com` will mess with the monitoring setup. See the image attached where the hostname changed to something that is not our primary host at 10:31.

screen shot 2015-05-08 at 07 23 07

rtomayko:
Tweaked the README copy to clarify the difference between restoring to an already configured instance vs. an instance that's never been configured in https://github.com/github/backup-utils/commit/b4b8bdb85c2beb8f3f37c2c8340b4cd63e272dbe.

I also created #128 to capture @Xeago's request for overriding configuration on restore.

Merging ...

candacebutler8879 (commented):
Ok

PR #100, opened 2015-02-02
Perform a full config run on already configured .340 and 2.x instances on restore

### LINKED ISSUE #95: ghe-restore should recreate cache directories on 11.10.3xx
Raising this on behalf of an Enterprise customer.

`ghe-restore` currently removes the `/data/repositories/__nodeload_archives__/`, `/data/repositories/__gitmon__/` and `/data/repositories/__render__/` cache directories when restoring to an 11.10.3[45]x appliance and doesn't recreate them forcing the administrator to remember to save the settings and/or perform a full configuration run after each restore or just before "going live" with the restore appliance.

In the case of continual backup & restore to another standby appliance, the settings are likely to have already been verified when the appliance was first created so the administrator is not likely to think to save them again after subsequent restores and will just disable maintenance mode.  This will result in things like Wikis, Pages and archive downloads not working.

lildude:
This PR changes a few things:

1.  It now calls `ghe-storage-prepare` directly instead of using `ghe-config-apply` on 2.x instances.  This is because `ghe-config-apply` will not call `ghe-storage-prepare` if there have been no changes to the settings, which would be the case in a repeated incremental restore to an already configured 2.x instance.  This is required for repeated restores of 2.x backups to another already configured 2.x instance and enables admins to disable maintenance mode and start using the instance without any further steps.

2.  It adds the equivalent for 11.10.34x restores thus allowing people to repeatedly restore an 11.10.34x backup to another 11.10.34x host and simply disable maintenance mode and everything will be running as expected.  This does have the side effect that this takes quite a bit longer to run than the equivalent on 2.x so I've added a nice message to warn of this.  This fixes #95.

This has the added benefit that this PR effectively introduces support for incremental 11.10.34x -> 2.x migrations as the user no longer needs to manually run `ghe-storage-prepare` after each restore on configured instance.


lildude:
Ooops, looks like I need to update the tests too. Will do that now.

rtomayko:
> It now calls ghe-storage-prepare directly instead of using ghe-config-apply on 2.x instances.

I think we may still need to call `ghe-config-apply` here or settings overwritten on restore won't be applied? Seems like ideally we'd want something like `ghe-config-apply --force-prepare-storage` but that obviously won't work for >= 2.0 < 2.2 versions :(

I wonder if we can fake `ghe-config-apply` by writing to some setting that'd cause it to run `ghe-storage-prepare` reliably. Let me poke around and there and see what I can come up with.

rtomayko:
One really gross way of forcing a `ghe-storage-prepare` run would be to append a blank line or comment to `/data/github/shared/env.d/99-instance.sh` right before running `ghe-config-apply`. Something like:

    echo "# storage prepare run forced" >> `/data/github/shared/env.d/99-instance.sh`
    ghe-config-apply

Another option would be to manually run it after if `ghe-config-apply` didn't already take care of it:

    if ! grep -q "running ghe-storage-prepare" /data/github/common/chef.log; then
        /usr/local/share/enterprise/ghe-storage-prepare
    fi

I have to say, I'm not in love with any of these approaches and we'll need to maintain whatever we do here within backup-utils for a while. I'd like to move away from calling any undocumented utilities outside of `/usr/local/bin`. I'm tempted to wait until we land `ghe-config-apply --force-prepare-storage`. We could probably get it in for the next 2.1.x patch release.

One nice thing is that `ghe-config-apply` doesn't check argv and fail on invalid arg currently, so we could probably plumb in the arg now. I'm also cool with landing the 340 logic.

lildude:
Great. I wasn't too comfortable with calling `ghe-storage-prepare` directly either but really wanted to get a solution out as this requirement is becoming a bit of a call generator.

I'll whip out the 2.x logic for the mo. We can come back to this when `ghe-config-apply --force-prepare-storage` lands in 2.x.

rtomayko:
Awesome, thanks! :sparkles:

lildude:
@rtomayko  We're all set on the GHE side of things for this to be merged in prep for the 2.0.7 and 2.1.3 releases.

rtomayko:
Sorry for the delay here.

Fixed a quick bug in the `if $instance_configured` conditional in 1461d6a. Check out the commit message for more. After that I wanted to add a test to verify that the config run is *not* performed when restoring to an unconfigured instance, which could lead to a lot of confusion during migrations. This failed initially due to the ghe-restore test suite not properly resetting fixture state between tests. Fixed that up and the test is passing now.

Feeling good about merging this. Thanks @lildude!

lildude:
Good catch. I'm happy for you to merge when happy.

PR #101, opened 2015-02-04
Add devscripts to debian's build-depends for checkbashisms

Xeago:


Xeago:
When building/testing inside a chroot, the package devscripts wasn't installed. It however is needed as a dependency for building this package.

We use [sbuild](https://wiki.debian.org/sbuild) to build using a single-use chroot, the package not being in the build dependencies causes a build failure when running checkbashisms.

Xeago:
Can this be merged? This is the equivalent of why we install devscripts in travis: https://github.com/github/backup-utils/blame/master/.travis.yml#L6

It is blocking us from upgrading our internal package - we don't want to maintain separate patches to our fork.

rtomayko:
Yep, super sorry. I missed this when it came through somehow.

PR #121, opened 2015-03-25
warn and prompt when replication is configured

buckelij:
This PR adds an additional check to warn and prompt when restoring to a server in a replication pair. Restoring to a server in a replication pair will interrupt replication, and requires that replication be reconfigured.

### DIFF
--- a/bin/ghe-restore
+++ b/bin/ghe-restore
@@ -97,9 +97,9 @@ elif [ "$GHE_VERSION_MAJOR" -ge 2 ]; then
     restore_settings=true
 fi
 
-# If you restore to a server in a replication pair
+# Figure out if this instance is in a replication pair
 if ghe-ssh "$GHE_HOSTNAME" -- "ghe-repl-status -r 2>/dev/null" \
-  | grep -E "replica|primary"; then
+  | grep -Eq "replica|primary"; then
   instance_configured=true
   echo "WARNING: Restoring to a server with replication enabled interrupts replication."
   echo "You will need to reconfigure replication after the restore completes."
--- a/bin/ghe-restore
+++ b/bin/ghe-restore
@@ -97,6 +97,14 @@ elif [ "$GHE_VERSION_MAJOR" -ge 2 ]; then
     restore_settings=true
 fi
 
+# If you restore to a server in a replication pair
+if ghe-ssh "$GHE_HOSTNAME" -- "ghe-repl-status -r 2>/dev/null" \
+  | grep -E "replica|primary"; then
+  instance_configured=true
+  echo "WARNING: Restoring to a server with replication enabled interrupts replication."
+  echo "You will need to reconfigure replication after the restore completes."
+fi
+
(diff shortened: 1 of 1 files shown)


rtomayko:
Oh nice :+1: Maybe we should abort here instead of warning. Writing into a primary's fs with replication enabled is probably never a good idea.

```
echo "ERROR: Refusing to restore to an instance with replication enabled."
echo "Please disable replication by ...
```

Unfortunately, I don't see where [`ghe-repl-teardown`](https://github.com/github/enterprise2/blob/master/vm_files/usr/local/bin/ghe-repl-teardown) is handling taking the primary out of replication mode so not sure how to suggest disabling.

Running `ghe-repl-stop` isn't enough. We really need to disable replication entirely or at least ensure that a full `ghe-repl-setup` is run after restoring.

rtomayko:
Anyway, :+1: on landing this for now. I'll open an issue on not being able to disable replication cleanly.

bin/ghe-restore
```diff
@@ -97,6 +97,14 @@ elif [ "$GHE_VERSION_MAJOR" -ge 2 ]; then
     restore_settings=true
 fi
 
+# If you restore to a server in a replication pair
+if ghe-ssh "$GHE_HOSTNAME" -- "ghe-repl-status -r 2>/dev/null" \
+  | grep -E "replica|primary"; then
```
rtomayko: You'll want to `>/dev/null` the `grep` output here or we'll get a matching line on stdout. Outside of that, I'm good to merge.

buckelij:
@rtomayko I added `-q` to grep.
`ghe-repl-status -r` only works on the latest releases of GitHub Enterprise. Is it worth directly inspecting  */etc/github/repl-state* on earlier versions?

rtomayko:
@buckelij Let's start here and add support for earlier versions if the need arises. Sorry, just noticed - can you redirect those error messages to stderr? Once that's in I think we can merge.

buckelij:
Should it go to stderr if we're not aborting?

bin/ghe-restore
```diff
@@ -97,6 +97,14 @@ elif [ "$GHE_VERSION_MAJOR" -ge 2 ]; then
     restore_settings=true
 fi
 
+# Figure out if this instance is in a replication pair
+if ghe-ssh "$GHE_HOSTNAME" -- "ghe-repl-status -r 2>/dev/null" \
+  | grep -Eq "replica|primary"; then
+  instance_configured=true
+  echo "WARNING: Restoring to a server with replication enabled interrupts replication."
+  echo "You will need to reconfigure replication after the restore completes."
```
tmm1: I would indent this second line so it's clear that it is part of the warning. (See example a few lines below).

buckelij: :+1:

tmm1:
Looks good to me. The other warnings in this file are not going to stderr, so I think it's fine to merge.

rtomayko:
Yep, thanks!

PR #129, opened 2015-05-13
Remove temporary 2.2 migration dir before restoring repos

lildude:
This PR removes the temporary /data/user/repositories-nw-backup directory that remains after successfully migrating the repository storage layout to the new format used on GHE 2.2.0 and later.

The removal of this directory allows for continual restoration to work on GHE 2.2.x.

/cc @rtomayko

### DIFF
--- a/bin/ghe-restore
+++ b/bin/ghe-restore
@@ -179,7 +179,7 @@ fi
 
 echo "Restoring Git repositories ..."
 # Remove temporary 2.2 storage migration directory if it exists
-echo "[ -d /data/user/repositories-nw-backup ]; then sudo rm -rf /data/user/repositories-nw-backup; fi" |
+echo "if [ -d /data/user/repositories-nw-backup ]; then sudo rm -rf /data/user/repositories-nw-backup; fi" |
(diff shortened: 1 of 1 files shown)


lildude:
Locally tests are failing.  I'll investigate further and update this PR.

lildude:
Fixed. It was a silly syntax error. All should be good now.

bin/ghe-restore
```diff
@@ -178,6 +178,9 @@ if [ "$GHE_VERSION_MAJOR" -ge 2 ]; then
 fi
 
 echo "Restoring Git repositories ..."
+# Remove temporary 2.2 storage migration directory if it exists
+echo "if [ -d /data/user/repositories-nw-backup ]; then sudo rm -rf /data/user/repositories-nw-backup; fi" |
```
rtomayko: We can axe the `if` here too. `rm -rf` will exit non-zero when the directory doesn't exist.

rtomayko: Actually, let's leave it. The `sudo` will cause problems on GHE < 2.0 and I don't think it's worth wrapping in a version check.

PR #131, opened 2015-05-21
Log backup/restore start/complete on remote instance

rtomayko:
Adds logging to `/var/log/syslog` on the remote GHE instance to both `ghe-backup` and `ghe-restore`. The logging is fairly simple and records only start and completion but that should be fine for now. This should work with remote syslog configurations since `logger(1)` is used as well.

Example messages:

```
May 21 11:29:30 ghe-io backup-utils: Starting backup from rtko.local in snapshot 20150521T075236
May 21 11:29:40 ghe-io backup-utils: Completed backup from rtko.local / snapshot 20150521T075236 successfully.
May 21 11:30:00 ghe-io backup-utils: Starting restore from rtko.local / snapshot 20150521T075236 ...
May 21 11:30:30 ghe-io backup-utils: Completed restore from rtko.local / snapshot 20150521T075236 ...
```

Internal ticket: https://github.com/github/enterprise2/issues/3805#issuecomment-103219349.

/cc @donal, @dbussink, @jatoben, @cjs

### DIFF
--- a/bin/ghe-backup
+++ b/bin/ghe-backup
@@ -67,6 +67,9 @@ echo "Starting backup of $GHE_HOSTNAME in snapshot $GHE_SNAPSHOT_TIMESTAMP"
 ghe_remote_version_required
 echo "$GHE_REMOTE_VERSION" > version
 
+# Log backup start message in /var/log/syslog on remote instance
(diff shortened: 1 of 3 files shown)


dbussink:
:+1:

mikemcquaid:
Seems sensible :+1:

PR #134, opened 2015-06-29
Remove hookshot backup/restore

tmm1:
The hookshot directory contains only log files of successful and failed webhooks. These are of limited value and not worth backing up and restoring (especially since they accumulate indefinitely on older instances). This will also reduce the time required to do a backup/restore or to upgrade to enterprise 2.x from a 11.10.x release.

cc @github/enterprise-support @github/enterprise-infrastructure @github/enterprise-releases @github/platform github/enterprise2#4504

### DIFF
--- a/share/github-backup-utils/ghe-backup-repositories-rsync
+++ b/share/github-backup-utils/ghe-backup-repositories-rsync
@@ -247,9 +247,9 @@ rsync_repository_data -H <<RULES
 RULES
 
 # Sync __special__ data directories, including the __alambic_assets__,
-# __hookshot__, and __purgatory__ directories. The __nodeload_archives__,
+# and __purgatory__ directories. The __nodeload_archives__,
 # __gitmon__, and __render__ directories are excludes since they act only as
-# caches.
+# caches. The __hookshot__ directory is also excluded since it only contains log files.
 #
 # Under v2.x and greater, only the special __purgatory__ directory remains under
 # /data/repositories. All other special user data directories have been moved under
@@ -260,6 +260,7 @@ rsync_repository_data <<RULES
 - /__nodeload_archives__/
 - /__gitmon__/
 - /__render__/
+- /__hookshot__/
 + /__*__/
 + /__*__/**
 + /info/
--- a/bin/ghe-backup
+++ b/bin/ghe-backup
@@ -123,10 +123,6 @@ if [ "$GHE_VERSION_MAJOR" -ge 2 ]; then
     echo "Backing up asset attachments ..."
     ghe-backup-userdata alambic_assets ||
     failures="$failures alambic_assets"
-
-    echo "Backing up hook deliveries ..."
-    ghe-backup-userdata hookshot ||
-    failures="$failures hookshot"
 fi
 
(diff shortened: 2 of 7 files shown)


share/github-backup-utils/ghe-backup-repositories-rsync
```diff
@@ -260,6 +260,7 @@ rsync_repository_data <<RULES
 - /__nodeload_archives__/
 - /__gitmon__/
 - /__render__/
+- /__hookshot__/
```
rtomayko: Don't think we want to remove this exclusion.

rtomayko:
Looks good to me with the exception of the __hookshot__ exclusion during repo backup.

share/github-backup-utils/ghe-backup-repositories-rsync
tmm1: I added an exclusion here, so we don't pull `/data/repositories/__hookshot__` during a backup from a 11.10.x release.

tmm1:
@rtomayko Can you elaborate why the hookshot shouldn't be skipped during backup?

rtomayko:
Sorry, I just read that backwards. Looks good. :+1:

dbussink:
Looks good to me :+1:

Xeago:
I'm slightly moved by the exclusion of the hookshot directory. We've had the need of investigating hook(shot) deliveries, and after a (controlled) failover we noticed that this is no longer possible. We opened a support request and were informed that we can copy /data/user/hookshot from the old primary.   
Were this not a controlled failover, we wouldn't have a chance to recover these deliveries for investigation, with this change.

tmm1:
I understand your concern, but large installations can generate gigabytes of hookshot logs which adds a significant strain to the backup/restore process. The backup-utils project was designed to backup user data, and currently ignores all log files on the appliance as they are considered transient- this now includes the hookshot logs. Delivery errors should be easy to reproduce with activity on the instance which will enqueue new webhooks.

Note also that although this PR has been merged to master, it has not made it to the stable release branch which we recommend customers use.

Xeago:
After a failover, especially an unexpected one, we don't expect a lot of activity while we establish that the instance is doing fine. As a matter of fact, we have our security group (in AWS) set up in such a way that the administrators can still access whilst we recover the service and verify it is working fine.  
Activity can also be very sporadic for different repositories, they might be based in a different timezone. Waiting for their activity really does not help troubleshooting and causes a lot of frustration on all continents.

I'm a bit confused that these are considered log files. whilst I don't understand the format, it appears to be a file per request. The results of the hookshot is viewable through the appliance itself, without any hackery or url-modiffying. Whereas any other log file is not viewable in such a way but is available of the log forwarding option - these hookshot 'logs' are essentially very well kept secrets by an appliance. 

Would keeping (and thus backing up) the last n days solve both our problems? Customers store m backups, thus having m+n days of hookshot information available to them.

lildude:
After internal discussions about the significant improvements in the handling of the hookshot logs on the appliance itself, it has been agreed to revert this pull request an reintroduce backing up of the hookshot logs.

PR #136, opened 2015-08-04
Revert "Remove hookshot backup/restore"

lildude:
Reverts github/backup-utils#134

Following discussions regarding keeping a backup of the hookshot logs, it has been decided to revert this change as the number of logs will be significantly fewer on later releases of GitHub Enterprise.

This does however reintroduce the severe delay when migrating from GitHub Enterprise 11.10.3xx to 2.x as GitHub Enterprise 11.10.3xx doesn't prune the log files like 2.2 does.

### DIFF
--- a/bin/ghe-backup
+++ b/bin/ghe-backup
@@ -123,6 +123,10 @@ if [ "$GHE_VERSION_MAJOR" -ge 2 ]; then
     echo "Backing up asset attachments ..."
     ghe-backup-userdata alambic_assets ||
     failures="$failures alambic_assets"
+
+    echo "Backing up hook deliveries ..."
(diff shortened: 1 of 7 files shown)


lildude:
On a side note, I plan to raise a new PR today which limits the number of hookshot logs we backup which will resolve the severe delay 11.10.3xx users will see when migrating.

lildude:
Looking into this further and we've never backed up the hookshot logs on 11.10.3xx instances so hookshot logs won't be affecting migrations from 11.10.34x to 2.x.

mcolyer:
:+1: to reverting this.

lildude:
Thanks @mcolyer. Merging.

Issue #137, opened 2015-08-06
"--link-dest arg does not exist"  message for empty hookshot log directory


cjs:
On the stable branch, if the backup host is using rsync > 3.0 and if the hookshot backup directory exists but is empty, the ``ghe-backup-userdata`` command fails with:

```
--link-dest arg does not exist: ../../current/hookshot
```

This shows up for users with GitHub Enterprise versions 2.2.5 and higher who don't run webhooks since the directory structure of the hookshot delivery logs changed. For 2.2.4 and below, there was a default ``logs`` subdirectory, but it's empty if there are no webhook deliveries for 2.2.5 and higher.

mkchandler:
I know this issue has been fixed in `master` via #138, but is there going to be a release that includes it soon? If not, is it safe for me to run the `master` branch on our backup server?

lildude:
A new release is likely to be made when GitHub Enterprise 2.4 is released later this month. There's no danger with using the current master in the meantime.

PR #138, opened 2015-08-06
Only use --link-dest when backing up populated directories

lildude:
Rsync 3.0.0 and later appears to have some optimizations which results in...

```
--link-dest arg does not exist: ../../current/hookshot
```

... errors when using the `--link-dest` argument is used with empty directories.  As there is nothing to hardlink to, we don't really need to use this argument so lets not use it.

This resolves https://github.com/github/backup-utils/issues/137 (and is dependent on https://github.com/github/backup-utils/issues/136 being merged first).

### DIFF
--- a/test/test-ghe-backup.sh
+++ b/test/test-ghe-backup.sh
@@ -251,3 +251,16 @@ begin_test "ghe-backup without manage-password file"
     [ ! -f "$GHE_DATA_DIR/current/manage-password" ]
 )
 end_test
+
+begin_test "ghe-backup emtpy hookshot directory"
+(
+  set -e
+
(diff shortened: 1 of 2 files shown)


mkchandler:
Any updates on when this will be merged and released?

test/test-ghe-backup.sh
```diff
@@ -251,3 +251,16 @@ begin_test "ghe-backup without manage-password file"
     [ ! -f "$GHE_DATA_DIR/current/manage-password" ]
 )
 end_test
+
+begin_test "ghe-backup emtpy hookshot directory"
```
mikemcquaid: `emtpy` -> `empty`

mikemcquaid:
@lildude Other than the typo: :+1: :shipit:

PR #140, opened 2015-08-28
Explain why backup-utils is useful even if you have the HA replica running

mattcantstop:
Last week during an Implementations engagement a customer didn't feel like the backup-utils tool was necessary because the data was being backed up by the High Availability replica instance. The DevOps person that I was working with asked me to write up an explanation as to why we needed it so he could get approval from his manager to provision another machine for the backups. This is that explanation. 

I thought it would best live at the bottom of this README file so we could point people to it easily. It also gives people a chance to internalize the setup with the different instances and know why they are setting up these different tools. 

CC: @nathos, @kylemacey and anyone on the maintainers team of backup-utils.


nathos:
@mattcantstop made some edits to match the style of the rest of the doc (and our help docs).

mattcantstop:
@nathos, I like the changes :sparkles:

README.md
```diff
@@ -231,6 +232,16 @@ most recent successful snapshot:
 Note: the `GHE_DATA_DIR` variable set in `backup.config` can be used to change
 the disk location where snapshots are written.
 
+### How do backup snapshots differ from the HA (High Availability) replica?
+The backup utilities and [HA replica](https://help.github.com/enterprise/admin/guides/installation/high-availability-cluster-configuration/) serve different yet complementary purposes as part of a recommended GitHub Enterprise deployment.
+
+##### The purpose of the High Avaibility replica
```
mikemcquaid: `Avaibility` typo

README.md
```diff
@@ -231,6 +232,16 @@ most recent successful snapshot:
 Note: the `GHE_DATA_DIR` variable set in `backup.config` can be used to change
 the disk location where snapshots are written.
 
+### How do backup snapshots differ from the HA (High Availability) replica?
+The backup utilities and [HA replica](https://help.github.com/enterprise/admin/guides/installation/high-availability-cluster-configuration/) serve different yet complementary purposes as part of a recommended GitHub Enterprise deployment.
```
mikemcquaid: `It is recommended that both backup utilities and a HA replica are used as part of a GitHub Enterprise deployment but they serve different roles.` :question:

README.md
```diff
@@ -231,6 +232,16 @@ most recent successful snapshot:
...
+
+##### The purpose of the High Avaibility replica
+The HA replica is a fully redundant secondary GitHub Enterprise instance, kept in sync with the primary instance via replication of all major datastores. This active/passive cluster configuration is designed to minimize service disruption in the event of hardware failure or major network outage affecting the primary instance. Because some forms of data corruption or loss may be replicated immediately from primary to replica, it is not a replacement for the backup utilities as part of your disaster recovery plan.
```
mikemcquaid: `The HA replica replicates all major datastore from the primary instance to minimize service disruption in even of a primary instance outage. As it may replicate data loss/corruption it is not a disaster recovery solution and the backup utilities should be used as well`.

README.md
```diff
@@ -231,6 +232,16 @@ most recent successful snapshot:
...
+
+##### The purpose of the backup utilities
+Backup utility snapshots can be used to restore a GitHub Enterprise instance to a prior state, or set up a new instance. The backup host does not require the overhead of running another GitHub Enterprise instance. The backup utilities copy repository data, along with full snapshots of all other pertinent data stores. Backup snapshots are date-stamped, making rollback to previous states possible.
```
mikemcquaid: `Backup utilities takes date-stamped snapshots of all major datastores. These snapshots are used to restore an instance to a prior state or set up a new instance without having another always-on GitHub Enterprise instance (like the HA replica).`

README.md
```diff
@@ -13,6 +13,7 @@ This repository includes backup and recovery utilities for [GitHub Enterprise][1
 - **[Using the backup and restore commands](#using-the-backup-and-restore-commands)**
 - **[Scheduling backups](#scheduling-backups)**
 - **[Backup snapshot file structure](#backup-snapshot-file-structure)**
+- **[How do backup snapshots differ from a HA (High Availability) replica?](#how-do-backup-snapshots-differ-from-a-ha-high-availability-replica)
```
mikemcquaid: `How does backup utilities differ from a HA (High Availability) replica?

mikemcquaid:
Looking pretty good so far. I've made some language suggestions; feel free to ignore them or roll-them in, up to you. A simpler way of explaining may be just to point out that an HA replica provides HA but backup utilities provides DR.

mattcantstop:
>  A simpler way of explaining may be just to point out that an HA replica provides HA but backup utilities provides DR.

That is super succinct @mikemcquaid. I like it. I will incorporate your changes with special attention called to that point. I really want it to hit home that, while technically the HA is a backup of the data, it is not a data backup. I'll make these changes today.

nathos:
@mattcantstop I don't even like to think of HA as "technically a backup". It's redundancy/disaster recovery/business continuity like @mikemcquaid mentioned.

Similar to the "[RAID is not a backup](http://stackoverflow.com/questions/19037143/why-is-raid-1-considered-bad-for-backup/21054888#21054888)" explanation:

> RAID is not a backup, it is hardware redundancy for the sole purpose of providing uninterrupted business continuity in the event of a hardware failure. Redundancy is not backup.
>
> RAID is not a backup.
> 
> A backup is a solution that allows you to revert to a known good copy of data in the event of data loss. Such losses can be a result of hardware failure, malicious intent, human error, or software bugs. Backup is not redundancy; it cannot provide uninterrupted business continuity.
> 
> RAID is not a backup.
>
> Backup and hardware redundancy are two completely separate disciplines with different purposes. Comparing RAID with Backup is like comparing dual tires on a truck with a spare tire.

RAID is not a backup.

mattcantstop:
> @mattcantstop I don't even like to think of HA as "technically a backup". It's redundancy/disaster recovery/business continuity like @mikemcquaid mentioned.

I was using the more colloquial use of the word. It is technically another copy of all your data.

mikemcquaid:
Looking pretty good :+1: . Still think it would be good to explicitly use the DR term, though and perhaps indicate that HA means High Availability (in case people see it elsewhere or are aware of the acronyms and not the words).

mattcantstop:
> Still think it would be good to explicitly use the DR term, though and perhaps indicate that HA means High Availability (in case people see it elsewhere or are aware of the acronyms and not the words).

Agreed @mikemcquaid. I changed "HA" to "High Availability" throughout and made specific mention of the backup utilities being a disaster recovery tool. 

Thanks for all the feedback :+1:

mikemcquaid:
Thanks @mattcantstop!

PR #141, opened 2015-09-01
Update current datetimestamp in directory structure example

nathos:
The "current" symlink in the directory structure wasn't pointing to the the latest snapshot.


mikemcquaid:
:+1:

PR #142, opened 2015-09-07
Fix rendering of "How does backup utilities..." link

lildude:
Fixes the rendering and link for the "How does backup utilities differ from a High Availability replica?" link.


michaeltwofish:
:+1:

PR #145, opened 2015-09-28
Move in progress detection to separate file with pid

### LINKED ISSUE #99: ghe-backup doesn't gracefully deal with failed backups
If a backup fails before finishing, causing the script to end prematurely, then subsequent runs will continue to fail as it'll think that a backup is still in progress.

````shell
$ ./bin/ghe-backup 
Error: backup of github.local already in progress in snapshot 20150129T135627. Aborting.
````

Rather than rely on the existence of a symlink to determine whether a backup is running, using a pid file and then confirming that the pid actually has a running process associated with it would be far, far more robust than the current method 

Relevant code- https://github.com/github/backup-utils/blob/master/bin/ghe-backup#L53

### PR DESCRIPTION
**Pull request by dbussink**
This changes the logic of detecting if a backup is in progress by writing out the snapshot and PID of the process making the backup. This improves the situation where a previous backup was killed in a non-clean way. Clean failures in the sense of a proper exit (also non 0 exit statuses) or using ctrl-c was already handled by traps.

This just removes the in-progress file if the process is no longer running after the backup. It then cleans up the in-progress file and lets the rest of the system clean up the failed backup.

Fixes #99

cc @github/enterprise-infrastructure

### DIFF
--- a/bin/ghe-backup
+++ b/bin/ghe-backup
@@ -56,9 +56,9 @@ trap 'cleanup' EXIT
 trap 'exit $?' INT # ^C always terminate
 
 if [ -h ../in-progress ]; then
-  echo "Error: in progress backup from previous version detected." 1>&2
+  echo "Error: detected a backup already in progress from a previous version of ghe-backup." 1>&2
   echo "If there is no backup in progress anymore, please remove" 1>&2
-  echo "the $GHE_DATA_DIR/in-progress symlink. This is only needed once." 1>&2
+  echo "the $GHE_DATA_DIR/in-progress symlink." 1>&2
   exit 1
 fi
 
--- a/bin/ghe-backup
+++ b/bin/ghe-backup
@@ -28,36 +28,55 @@ touch "incomplete"
 GHE_MAINTENANCE_MODE_ENABLED=false
 
 # To prevent multiple backup runs happening at the same time, we create a
-# in-progress symlink pointing to the snapshot directory. This will fail if
-# another backup is already in progress, giving us a form of locking.
+# in-progress file with the timestamp and pid of the backup process,
+# giving us a form of locking.
 #
-# Set up a trap to remove the in-progress symlink if we exit for any reason but
-# verify that we own the in-progress symlink before doing so.
+# Set up a trap to remove the in-progress file if we exit for any reason but
+# verify that we are the same process before doing so.
 #
 # The cleanup trap also handles disabling maintenance mode on the appliance if
 # it was automatically enabled.
 cleanup () {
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: bin/ghe-backup [hunk 1]
```diff
@@ -28,37 +28,56 @@ touch "incomplete"
...
 trap 'exit $?' INT # ^C always terminate
 
-# Mark the snapshot as in-progress by creating the symlink. If this fails, it
-# means another ghe-backup run is already in progress and we should exit.
-# NOTE: The -n argument to ln is non-POSIX but widely supported.
-if ! ln -sn "$GHE_SNAPSHOT_TIMESTAMP" ../in-progress 2>/dev/null; then
-    snapshot="$(readlink ../in-progress)"
-    echo "Error: backup of $GHE_HOSTNAME already in progress in snapshot $snapshot. Aborting." 1>&2
+if [ -h ../in-progress ]; then
+  echo "Error: in progress backup from previous version detected." 1>&2
```
michaeltwofish: I think we can improve the error message. Was there a reason to drop `$GHE_HOSTNAME` from it?

How about:

> Error: detected a backup of $GHE_HOSTNAME already in progress.

bin/ghe-backup
```diff
@@ -28,37 +28,56 @@ touch "incomplete"
...
+  echo "If there is no backup in progress anymore, please remove" 1>&2
+  echo "the $GHE_DATA_DIR/in-progress symlink. This is only needed once." 1>&2
```
michaeltwofish: It's not clear what is only needed once. Is it removing the file? And it's not a symlink any more.

michaeltwofish: Ah, I missed that this was testing for a symlink, so "previous version" and "only needed once" make more sense to me now, and why it's tricky to write a good error message.

```
echo "Error: detected a backup already in progress from a previous version of ghe-backup." 1>&2
echo "If there is no backup in progress anymore, please remove" 1>&2
echo "the $GHE_DATA_DIR/in-progress symlink." 1>&2
```

I don't think it's necessary to say it's only needed once if we're specific about the version of ghe-backup.

michaeltwofish:
I've left some comments on the error copy, but the functional changes make sense to me.

bin/ghe-backup
dbussink: Comment seems outdated based on the additional feedback, so going with the last mentioned suggestion.

dbussink:
@michaeltwofish Updated the wording with your suggestion, thanks!

lildude:
:+1: looks good to me.

Issue #149, opened 2015-11-23
S3 backups should backup more than just the most recent backup (current)


akarnani:
The S3 scripts only upload/download the current directory.  The upload intends to overwrite all previous data.  However we would find it useful to have the same file structure in S3 as on disk so that we can roll back to arbitrary backups.

We don't trust EBS enough for it to be our sole repository for backups.

brntbeer:
hey @akarnani you should take a look at #150. The new command we're looking to add is `sync` just like yours. 

Since that change uploads just the `current` folder. There will be a change we'll be looking to add afterwards that will probably allow you to pass in a variable to say `--all` and sync the entire data folder.

akarnani:
Seems pretty reasonable to me.  Looking forward to the `--all` change!

brntbeer:
@akarnani totally. good call on using `sync` too (@jasonmassey found it on his own from past experiences too, so it's just me here who's left being new!), it runs **SO** much faster.

jasonmassey:
Closing, should be fixed with #150

Issue #148, opened 2015-11-23
S3 backups do not currently work


akarnani:
The invocation of s3cmd is currently invalid.  Version `1.5.1.2` of `s3cmd` complains `ERROR: Parameter problem: Expecting S3 URI with just the bucket name set instead of 's3://'` from the invocation.

brntbeer:
@akarnani Sorry for the delay, I just ran through the `ghe-s3-backup` command and got a similar issue.

However, you may want to get a newer version of `s3cmd`, this should have been fixed in version `1.6.0`. The commit in question may be https://github.com/s3tools/s3cmd/commit/49966de886f9cebef5dd31bcf252311802f4c641. See it for more details.

Secondly, I also got this error when I didn't have `$GHE_S3_BUCKET` set, @jasonmassey and I are doing some work over in #150 to add that as an item in `backup.config`.

I'm going to close this down, since it may be an issue with version of s3cmd or variable

Issue #92, opened 2015-01-16
Does backup to s3 work?


xbeta:
I saw this line that uses `s3cmd` to `put` to s3, but we all know S3 as an Object Storage don't support empty folders, which some bare git repo has, how do we address that?

https://github.com/github/backup-utils/blob/master/share/github-backup-utils/ghe-s3-backup#L26

danjenkins:
I'm interested to know this too. It seems like a feature that isn't documented anywhere and that makes me nervous about using it. Can you give any info on whether I can use this?

pholleran:
I am at a DevOps Day event in Detroit and have had 3 different people ask about this capability.

rubiojr:
@xbeta @pholleran @danjenkins sorry for the late reply here, see https://github.com/github/backup-utils/issues/152#issuecomment-184677656

lildude:
It's worth pointing out that whilst there is _some_ S3 functionality in the scripts in backup-utils, it's not supported and doesn't work correctly, primarily because S3 doesn't support symlinks and each repository contains a symlink that would be lost when you restore your backup.

SEJeff:
@lildude I guess it could simply tar the backups then, that way preserving symlinks.

lildude:
> @lildude I guess it could simply tar the backups then, that way preserving symlinks.

Yup. The comment I made at https://github.com/github/backup-utils/issues/259#issuecomment-255329898 may help with making the tar more rsync friendly too.

fcostanzo-ha:
I wondered about using s3fs.

lildude:
> I wondered about using s3fs.

This could work, though this would be managed entirely _outside_ of Backup Utilities. 

It appears s3fs supports symlinks so that part should be ok but it doesn't support hard links, which will detect and prevent you from using the filesystem anyway. 

If Backup Utilities  doesn't correctly detect the lack of hardlink support and it starts using the filesystem, you're not going to get any of the space saving benefits and will end up uploading your entire backup to S3 every time you make a backup. This is going to get very very expensive very quickly.

The other [limitations](https://github.com/s3fs-fuse/s3fs-fuse#limitations) could pose problems with speed and cost too so it's probably not the best idea to use.

lildude:
Closing this as adding native S3 support is not on our current roadmap and not likely to be for some time.

If and when S3 or any other object-based storage support is added, this will be clearly documented in the release notes.

PR #150, opened 2015-12-03
Update S3 backup script

jasonmassey:
It seems like the S3 backup script needs some :heart: (see #92, #148 and #149)

Using s3cmd's "sync" seems like a good place to start. By default an md5 checksum and file size of each file is compared, and anything that has changed in the specified directory will be transferred to S3. Using the "--delete-removed" flag will keep the S3 backup mirroring the number of backups currently created and kept by backup-utils. Removing this flag will keep a full backup history in S3.

### DIFF
--- a/share/github-backup-utils/ghe-s3-backup
+++ b/share/github-backup-utils/ghe-s3-backup
@@ -22,5 +22,4 @@ ghe-backup
 s3cmd mb s3://$GHE_S3_BUCKET
 
 # Upload to S3.
-cd "$GHE_DATA_DIR"/current
-s3cmd sync --delete-removed --follow-symlinks * s3://$GHE_S3_BUCKET
+s3cmd sync --delete-removed --follow-symlinks "$GHE_DATA_DIR"/current/ s3://$GHE_S3_BUCKET
--- a/share/github-backup-utils/ghe-s3-backup
+++ b/share/github-backup-utils/ghe-s3-backup
@@ -23,4 +23,4 @@ s3cmd mb s3://$GHE_S3_BUCKET
 
 # Upload to S3.
 cd "$GHE_DATA_DIR"/current
-s3cmd --preserve put * s3://$GHE_S3_BUCKET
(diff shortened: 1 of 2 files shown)


jasonmassey:
@brntbeer using sync should take care of the "only the most recent backup" problem (#149). Let's give this a whirl and see what else we need to tweak. It seems like s3cmd might be picky about paths, so we may need to change from wildcards to explicit paths.

share/github-backup-utils/ghe-s3-backup
```diff
@@ -22,5 +22,4 @@ ghe-backup
 s3cmd mb s3://$GHE_S3_BUCKET
 
 # Upload to S3.
-cd "$GHE_DATA_DIR"/current
-s3cmd --preserve put * s3://$GHE_S3_BUCKET
+s3cmd sync --delete-removed --follow-symlinks "$GHE_DATA_DIR"/current/ s3://$GHE_S3_BUCKET
```
brntbeer: Since this command is kind of foreign and the command line flags are certainly unique, could you add a comment section above this to describe what some of these flags are doing?

sbryant:
I'm :cool: with this change. When it's :green_apple: feel free to merge. :+1:

brntbeer:
Also just tested this by running it twice, on the second run it finished after doing a comparison (finished in like 30 seconds). :sparkles:

brntbeer:
also wanting to cc @Xeago since you worked on this file last. Whatya think of these changes?

Xeago:
My work on this file is really just a structural change in the repository. History, prior to this includes commit 0c40b70c3ddce2dc293b63f407b04fcac31167b4 & 8d1beea324c6b5cb34443b10c4760936a6d7e1fd, both by @rtomayko. This last commit is also a structural change from `/scripts/` to `/bin/`. That said, I really do not feel qualified to advice on this pull request :)

The one thing that stands out a bit to me is that this is a shared library script? From the looks of it, this is intended to be invoked by the user or cron. I think it should be in `bin`, having it there will also install

PR #154, opened 2016-01-21
update new S3 backup/restore method 

carol-hsu:
Hi guys,

I would like to add this feature about saving data on AWS S3, following items are newly added:

- **do not backup GHE credential:** 
<br />I think it is not secure enough to put credential on AWS or in the data transfer stream. So users have to put the credential in backup host and provide full path in backup.config
- **compress data as a tarball before sending it to s3:**
<br />Backup data in tree structure of directory would not show directly in AWS web console. And we can have multiple snapshots in the same bucket.
- **enable S3's server side encryption with customer key(SSE-C):**
<br />To encrypt the GHE tarball on S3: the encrypted file can only be accessed through API(or AWS CLI) with accurate key.
<br /> Key is based on AES-256, users can generate it by command openssl: <br />copy the value of "iv" as a key file

> $ openssl enc -aes-256-cbc -k **PASSPHRASE** -P -md md5
salt=C1795304E61A6A34
key=7910918057FD9DE2BED0C68D1E41AC93A5379FD8FA1C01561606EED428D7CD4B
iv =C01231C0916845E3BB9B387D9089759B

To achieve SSE-C, I remove all **s3cmd** to **aws s3/s3api**. It will also be more convenient to run backup-utils on an AWS instance: no need to install "s3cmd".

I think this pull request could be an independent branch.
Any suggestion is welcome!

Thanks

Ref: [AWS S3's server side encryption with customer key](http://docs.aws.amazon.com/AmazonS3/latest/dev/ServerSideEncryptionCustomerKeys.html)


rubiojr:
Hey @carol-hsu, thanks. My apologies for the late reply.

Please give us a couple of days, currently looking into this.

carol-hsu:
Hi @rubiojr 
Thanks for your response, sorry that I failed to pass the test. 
I think it is because I remove the credential part.
Have a nice weekend!

rubiojr:
@carol-hsu the experimental S3 support [was removed recently](https://github.com/github/backup-utils/pull/167) because it was broken and causing trouble. We haven't been able to work on replacing it with a better implementation but it's definitely something I'd like to have at some point.

PR #155, opened 2016-02-03
Add cert to -c flag description

Fermentedbeats:
A customer inquired if their cert would carry over to the target instance when restoring.  We tested and it looks like a 'yes.'  Would love to spell this out in the docs for future reference.

Thanks and send any questions back my way.

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,7 +95,7 @@ download the most recent GitHub Enterprise version.
 
 ### Migrating from GitHub Enterprise v11.10.34x to v2.0, or v2.1
 
-If you are migrating from GitHub Enterprise version 11.10.34x to 2.0 or 2.1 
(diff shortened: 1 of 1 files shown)


rubiojr:
Thanks @Fermentedbeats, I think it makes sense to clarify docs a bit.

To clarify things a bit more, the certificates are embedded in the configuration (or settings) file, so restoring settings restores the certificates. Restoring the certificate doesn't need the `-c` flag, that is, the certificate will always be restored to a clean instance without having to use `-c`, just like the rest of the settings.

README.md
```diff
@@ -158,7 +158,7 @@ enable when output is logged to a file.
 When restoring to an already configured GHE instance, settings and license data
```
rubiojr: You may want to mention it here too.

Fermentedbeats: :heart: - thanks, @rubiojr!  Just added it :octocat:

PR #159, opened 2016-02-12
Add explicit instruction about adding backup host SSH key to target for `ghe-restore`

mellybess:
I received customer feedback that this was not explicit in the backup procedures and it caused some confusion during a test restore. Would :heart: to save future customers the same confusion when they are getting started.

Thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -109,7 +109,8 @@ After the initial backup, use the following commands:
  - The `ghe-backup` command creates incremental snapshots of repository data,
    along with full snapshots of all other pertinent data stores.
  - The `ghe-restore` command restores snapshots to the same or separate GitHub
-   Enterprise appliance. You must add the backup host's SSH key to the target GitHub Enterprise appliance before using this command.
+   Enterprise appliance. You must add the backup host's SSH key to the target
+    GitHub Enterprise appliance before using this command.
 
 ##### Example backup and restore usage
 
--- a/README.md
+++ b/README.md
@@ -109,7 +109,7 @@ After the initial backup, use the following commands:
  - The `ghe-backup` command creates incremental snapshots of repository data,
    along with full snapshots of all other pertinent data stores.
  - The `ghe-restore` command restores snapshots to the same or separate GitHub
-   Enterprise appliance.
+   Enterprise appliance. You must add the backup host's SSH key to the target GitHub Enterprise appliance before using this command.
 
 ##### Example backup and restore usage
 


sbryant:
Looks sane to me :+1: 

/cc @dbussink to see if he has any feels here

README.md
```diff
@@ -109,7 +109,7 @@ After the initial backup, use the following commands:
  - The `ghe-backup` command creates incremental snapshots of repository data,
    along with full snapshots of all other pertinent data stores.
  - The `ghe-restore` command restores snapshots to the same or separate GitHub
-   Enterprise appliance.
+   Enterprise appliance. You must add the backup host's SSH key to the target GitHub Enterprise appliance before using this command.
```
dbussink: We do line wrapping for the rest of readme, so we should apply that here too.

mellybess: It looks like the text is wrapping in the [rendered view](https://github.com/mellybess/backup-utils/blob/master/README.md#using-the-backup-and-restore-commands), is there something else I need to do?

rubiojr:
@mellybess the "Getting started" section mentions:

> Run bin/ghe-host-check to verify SSH connectivity with the GitHub appliance.

Which will result in:

```
[951][rubiojr@octox] ./bin/ghe-host-check
Permission denied (publickey).
Error: ssh connection with 'ghe-cluster.com' failed
Note that your SSH key needs to be setup on ghe-cluster.com as described in:
```

If the ssh key has not been added. Trying to run `ghe-backup` or `ghe-restore` will have the same effect.

The README also mentions:

> Add the backup host's SSH key to the GitHub appliance as an Authorized SSH key. See Adding an SSH key for shell access for instructions.

Same "Getting started" section, step 3.

Should be enough, IMO. Am I missing something?

README.md
dbussink: Yeah, it should also be wrapped in the raw view like the rest of the file.

mellybess:
> Should be enough, IMO. Am I missing something?

@rubiojr I had the same thought, but [we got feedback](https://halp.githubapp.com/discussions/3d73777ebee011e5917f1bd093207a11?page=1#C51) from the customer that no mention was made that the backup host SSH key had to be added to the target instance when restoring to a newly provisioned instance (in this case for testing) and this caused confusion.

README.md
mellybess: @dbussink Not sure if adding a line break is what you were asking me to do, if not let me know and I'll try to correct it.

sbryant:
> Should be enough, IMO. Am I missing something?

@mellybess's contribution is a good one, the README focuses more on setting up backups and is less clear on restores. At least that's how I felt after giving the entire document.

rubiojr:
> @mellybess's contribution is a good one, the README focuses more on setting up backups and is less clear on restores.

:+1: sounds reasonable to me.

README.md
rubiojr: @mellybess yeah, manual line break :+1:

rubiojr:
Thanks @mellybess!

PR #162, opened 2016-02-18
Update ghe-host-check to detect extra port 22 error

snh:
In some situations where port 22 is blocked, `ssh` will return a `port 22: No route to host` error instead of a `port 22: Connection refused` error. This PR adds handling to `bin/ghe-host-check` for this alternative error.

### DIFF
--- a/bin/ghe-host-check
+++ b/bin/ghe-host-check
@@ -37,7 +37,7 @@ set -e
 if [ $rc -ne 0 ]; then
     case $rc in
         255)
-            if echo "$output" | grep -i "port 22: connection refused\|Connection timed out during banner exchange" >/dev/null; then
+            if echo "$output" | grep -i "port 22: connection refused\|port 22: no route to host\|Connection timed out during banner exchange" >/dev/null; then
                 exec "bin/$(basename $0)" "$hostname:122"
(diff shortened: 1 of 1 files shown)


rubiojr:
@snh looks good to me but to double check, I'm trying to figure out when that'd be the case. Firewall? trying to reach a non-routable/private network? Do you know how to reproduce this one? Are you backing up or restoring a cluster or a single VM?

snh:
The error can be a bit misleading as it doesn't necessarily mean the backup host has no route to the GHE appliance.

It is usually caused by a firewall in the path rejecting the connection with a _ICMP Admin Prohibited_ response. The specific use case this PR solves is those where port 22 is blocked in this way but port 122 isn't. At the moment the backup fails even though the GHE appliance might be a 2.x release that only requires port 122 be accessible for backups, this change allows it to succeed.

The easiest way to reproduce is to add the following `iptables` rule on the backup server which is running `backup-utils`

```
iptables -I OUTPUT -p tcp -d <gheinstance> --dport 22 -j REJECT --reject-with icmp-admin-prohibited
```

rubiojr:
Thanks @snh!

PR #166, opened 2016-02-22
Prevent multiple restores from running at the same time

snh:
It is possible for multiple restores to run simultaneously, which can and likely will result in errors.

A situation where this might occur is where a restore is scheduled to occur automatically for backup validation or DR purposes and for one reason or another the previous restore does not finish before the next one runs.

This update leverages the existing `$GHE_REMOTE_DATA_USER_DIR/common/ghe-restore-status` file that is used to lock out the _Management Console_ during restores in GitHub Enterprise 2.0.0 and above.

If running GitHub Enterprise 2.0.0 and above, and `$GHE_REMOTE_DATA_USER_DIR/common/ghe-restore-status` contains `restoring`, the `bin/ghe-restore` script will abort.

/cc @michaeltwofish

### DIFF
--- a/bin/ghe-restore
+++ b/bin/ghe-restore
@@ -146,10 +146,6 @@ fi
 echo "Starting restore of $GHE_HOSTNAME from snapshot $GHE_RESTORE_SNAPSHOT"
 ghe_remote_logger "Starting restore from $(hostname) / snapshot $GHE_SNAPSHOT_TIMESTAMP ..."
 
-# Update remote restore state file and setup failure trap
-trap "update_restore_status failed" EXIT
-update_restore_status "restoring"
-
 # Verify the host has been fully configured at least once if when running
 # against v11.10.x appliances and the -c option wasn't specified.
 if [ "$GHE_VERSION_MAJOR" -le 1 ] && ! $restore_settings && ! $instance_configured; then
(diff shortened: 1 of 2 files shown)


bin/ghe-restore
```diff
@@ -176,6 +172,16 @@ if $instance_configured; then
     fi
 fi
 
+# Make sure the appliance doesn't already have a restore underway
+if [ "$GHE_VERSION_MAJOR" -ge 2 ] && ghe-ssh "$GHE_HOSTNAME" -- "sudo grep -q 'restoring' '$GHE_REMOTE_DATA_USER_DIR/common/ghe-restore-status' 2>/dev/null"; then
```
rubiojr: Minor style nit but single quotes in:

```
sudo grep -q 'restoring' '$GHE_REMOTE_DATA_USER_DIR/common/ghe-restore-status'
```

are not required, AFAICT.

rubiojr:
Thanks @snh, this looks good to me. /cc @github/backup-utils

bin/ghe-restore
snh: You are indeed correct, have tidied this up. Thanks!

lildude:
LGMT too.

gnawhleinad:
:ship:!

PR #170, opened 2016-02-24
Fix remote logging output to correctly show snapshot being restored

snh:
This small PR fixes the remote logging output so that it correctly shows the snapshot being used for the restore. It currently incorrectly outputs a timestamp of when the restore started rather than that of the snapshot being used.

Also incorporates a couple of minor tweaks to the wording of the remote logging messages to make them consistent.

/cc @github/backup-utils

### DIFF
--- a/bin/ghe-backup
+++ b/bin/ghe-backup
@@ -213,10 +213,10 @@ echo "Completed backup of $GHE_HOSTNAME in snapshot $GHE_SNAPSHOT_TIMESTAMP at $
 
 # Exit non-zero and list the steps that failed.
 if [ -z "$failures" ]; then
-    ghe_remote_logger "Completed backup from $(hostname) in snapshot $GHE_SNAPSHOT_TIMESTAMP successfully."
+    ghe_remote_logger "Completed backup from $(hostname) / snapshot $GHE_SNAPSHOT_TIMESTAMP successfully."
 else
     steps="$(echo $failures | sed 's/ /, /g')"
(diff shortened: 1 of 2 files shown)


bin/ghe-backup
```diff
@@ -94,7 +94,7 @@ if ghe-ssh "$GHE_HOSTNAME" -- \
 fi
 
 # Log backup start message in /var/log/syslog on remote instance
-ghe_remote_logger "Starting backup from $(hostname) in snapshot $GHE_SNAPSHOT_TIMESTAMP ..."
+ghe_remote_logger "Starting backup from $(hostname) / snapshot $GHE_SNAPSHOT_TIMESTAMP ..."
```
michaeltwofish: I think it's clearer with "in". With the slash it looks like backing up _from_ hostname/snapshot.

michaeltwofish:
Clearly a bug, so :+1:. Can we log failure as well?

rubiojr:
Replacing `GHE_SNAPSHOT_TIMESTAMP` with `GHE_RESTORE_SNAPSHOT` sounds good to me, but the other changes to replace `/` with `in` could break log parsers. Can we skip those? Looks like a cosmetic issue only.

snh:
> Replacing GHE_SNAPSHOT_TIMESTAMP with GHE_RESTORE_SNAPSHOT sounds good to me, but the other changes to replace / with in could break log parsers. Can we skip those? Looks like a cosmetic issue only.

Sure, sounds like a good approach for now. I have revered these to their original format.

gnawhleinad:
LGTM, :ship:!

PR #172, opened 2016-02-25
fall back to system /etc/github-backup-utils/ghe-backup-config if none found in share

austin987:


gnawhleinad:
We should be doing this in [`ghe-backup-config#L46-L56`](https://github.com/github/backup-utils/blob/master/share/github-backup-utils/ghe-backup-config#L46-L56). Is this not working for you?

rubiojr:
@gnawhleinad yeah, that should be the case.

@austin987 closing this one but, please, let us know if that's not working for you for some reason.

austin987:
Yeah, it appears to be working, thanks. I noticed it while debugging something else.

PR #173, opened 2016-02-25
Clustering: ghe-restore fixes

rubiojr:
More robust restores when restoring gists, pages and storage data into a
cluster.

/cc @github/backup-utils


gnawhleinad:
:ship:!

PR #174, opened 2016-02-25
Revert "Prevent multiple restores from running at the same time"

rubiojr:
Reverts github/backup-utils#166

If you abort the restore (i.e. `ctrl-c`), this will leave the `ghe-restore-status` file around, causing subsequent restores to fail.

We should probably remove the status file from the trap in that case but reverting this for now.


rubiojr:
Also doesn't take clustering into account, since you may land in a different server where the file is not present.

rubiojr:
@snh sorry I overlooked this the other day, wanna work on a new PR?

gnawhleinad:
:+1:!

snh:
> @snh sorry I missed this the other day, wanna work on a new PR?

No worries! I will work on a new PR :+1:

PR #175, opened 2016-02-25
Ask for index name when using Elasticsearch _cat API

jatoben:
Elasticsearch 1.2 didn't return the index open/closed state in the `_cat` API response, so `ghe-backup-es-audit-log` would fail to get the index names correctly on GitHub Enterprise 2.2 and earlier. This would result in `ElasticsearchIllegalArgumentException[Malformed scrollId []]` failures during the backup.

Asking specifically for the index name avoids this issue.

### DIFF
--- a/share/github-backup-utils/ghe-backup-es-audit-log
+++ b/share/github-backup-utils/ghe-backup-es-audit-log
@@ -19,7 +19,11 @@ ghe_remote_version_required "$host"
 # Make sure root backup dir exists if this is the first run
 mkdir -p "$GHE_SNAPSHOT_DIR/audit-log"
 
-indices=$(ghe-ssh "$host" 'curl -s "localhost:9201/_cat/indices/audit_log*?h=index"') || true
+if ! indices=$(ghe-ssh "$host" 'curl -s "localhost:9201/_cat/indices/audit_log*?h=index"'); then
+  echo "Error: failed to retrieve audit log indices." 1>&2
+  exit 1
+fi
+
 current_index=audit_log-$(ghe-ssh "$host" 'date +"%Y-%m"')
 
 for index in $indices; do
--- a/share/github-backup-utils/ghe-backup-es-audit-log
+++ b/share/github-backup-utils/ghe-backup-es-audit-log
@@ -19,7 +19,13 @@ ghe_remote_version_required "$host"
 # Make sure root backup dir exists if this is the first run
 mkdir -p "$GHE_SNAPSHOT_DIR/audit-log"
 
-if ! indices=$(ghe-ssh "$host" 'curl -s "localhost:9201/_cat/indices/audit_log*?h=index"'); then
+if [ $GHE_VERSION_MAJOR -ge 2 ] && [ $GHE_VERSION_MINOR -ge 2 ]; then
+  es_port=9201
+else
+  es_port=9200
+fi
+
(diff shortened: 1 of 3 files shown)


jatoben:
Oof, the test failures were caused by the `curl` stub. It echoes its arguments back, but the real thing would either return a list of audit logs, or an empty list on 11.10.3xx. This was masked before because the output was piped thru `cut`.

I added an ugly workaround for now, but I'd love any better suggestions.

/cc @github/backup-utils @rubiojr @dbussink @github/enterprise-support

share/github-backup-utils/ghe-backup-es-audit-log
```diff
@@ -19,7 +19,7 @@ ghe_remote_version_required "$host"
 # Make sure root backup dir exists if this is the first run
 mkdir -p "$GHE_SNAPSHOT_DIR/audit-log"
 
-indices=$(ghe-ssh "$host" 'curl -s "localhost:9201/_cat/indices/audit_log*"' | cut -d ' ' -f 3)
+indices=$(ghe-ssh "$host" 'curl -s "localhost:9201/_cat/indices/audit_log*?h=index"') || true
```
rubiojr: I'd prefer better error handling here and warn if we can't backup audit logs, instead of ignoring the error.

rubiojr:
> I added an ugly workaround for now, but I'd love any better suggestions.

:+1: to the workaround for now, integration tests will take care of making sure we're  backing up audit logs properly.

rubiojr:
@jatoben pushed some minor changes to guard against failure. Let me know what you think.

rubiojr:
> This will always fail on pre-2.2.0 because haproxy didn't listen on port 9201. We could add a check to use port 9200 on earlier versions, but we'd still have to handle failure on 11.10.3xx, since Elasticsearch 0.90 didn't provide the _cat API.

Ugh, good catch. I'd be in favor of fixing this correctly for 2.0 and 2.1 using version detection and warn (instead of `exit 1`) for 11.10.XXX.

> Would it be better to just skip this step when backing up pre-2.2 appliances?

No, I think it's doable without adding a lot of complexity.

jatoben:
I think this should be ok now, I tested it manually against 11.10.348 and the entire 2.x series.

Microchipen:
Please stop sending them to me


Bjørn Ekman
25. feb. 2016 23:09 skrev "Ben Gollmer" <notifications@github.com>:

dbussink:
@Microchipen Looks like you are getting these messages because you are subscribed to this repository on GitHub.

Microchipen:
How do I stpp that?
4. mars 2016 12:08 skrev "Dirkjan Bussink" <notifications@github.com>:

dbussink:
https://help.github.com/articles/unwatching-repositories/ has the documentation for how to do that.

Microchipen:
Can't find it :/
4. mars 2016 12:13 skrev "Dirkjan Bussink" <notifications@github.com>:

lildude:
> Can't find it :/

I've checked the link @dbussink referenced and it's definitely there and responding and provides all the details on how to unwatch a repository.

From https://github.com/github/backup-utils/watchers?page=2 we can see you are currently a watcher of this repository so unwatching will stop the notifications.

You can alternatively modify your notification settings as detailed in https://help.github.com/articles/managing-notification-delivery-methods/

If you're still having problems, please contact the Support team using the form at https://github.com/support as this isn't really the best place to discuss your notification settings etc.

Issue #160, opened 2016-02-13
“unknown operator” error from ghe-backup-es-audit-log


andrewdotn:
I don’t know if you’d consider this issue to be under the scope of https://github.com/github/backup-utils/issues/85, but I ran into an issue upgrading the backup tools today—

On our backup server, which runs SmartOS so that we can take advantage of ZFS, `/bin/sh` is actually ksh93, which doesn’t like this line:

    if [ -f $GHE_DATA_DIR/current/audit-log/$index.gz -a $index \< $current_index ]; then

With `\<` it complains about an unknown operator; with plain `<` it tries to read from `$current_index` as a file.

I worked around this by changing the shebang to `#!/bin/bash`, and am not sure if there’s a way to get it to work in ksh93 on SmartOS.

rubiojr:
@andrewdotn some of the scripts are already using `#!/usr/bin/env bash` to avoid compatibility issues (assuming you have bash installed). We'll probably end up doing the same for all the scripts in backup-utils. Will that work for you?

/cc @github/backup-utils

andrewdotn:
Yup, that would work just fine.

janorn:
This also affects solaris. Bash would solve it.

snh:
Fixed by #177

PR #176, opened 2016-02-25
Prevent multiple restores

snh:
New version of #166 that traps SIGINT, SIGTERM, SIGHUP, and SIGQUIT to ensure that future restores aren't incorrectly blocked by a stale `/data/user/common/ghe-restore-status` file that incorrectly indicates a restore is still in progress.

I would appreciate some advice on how best to handle a cluster scenario... At the moment this PR only checks the host specified as the restore host (either via `GHE_RESTORE_HOST` in the config file, or specified at the command line when running `ghe-restore`). 

This is in part because this is the only place we currently maintain a `/data/user/common/ghe-restore-status` file.

In its current form this will prevent multiple restores in a cluster environment where the same host is specified as the restore host, but not where different hosts are specified that are part of the same cluster. For automated scenarios, which I believe is likely to be the most common scenario where multiple restores run at the same time, I imagine the same host is generally used.

### DIFF
--- a/bin/ghe-restore
+++ b/bin/ghe-restore
@@ -146,10 +146,6 @@ fi
 echo "Starting restore of $GHE_HOSTNAME from snapshot $GHE_RESTORE_SNAPSHOT"
 ghe_remote_logger "Starting restore from $(hostname) / snapshot $GHE_RESTORE_SNAPSHOT ..."
 
-# Update remote restore state file and setup failure trap
-trap "update_restore_status failed" EXIT
-update_restore_status "restoring"
-
 # Verify the host has been fully configured at least once if when running
 # against v11.10.x appliances and the -c option wasn't specified.
 if [ "$GHE_VERSION_MAJOR" -le 1 ] && ! $restore_settings && ! $instance_configured; then
@@ -167,6 +163,16 @@ if [ "$GHE_VERSION_MAJOR" -le 1 ] && [ "$GHE_BACKUP_STRATEGY" = "rsync" ]; then
         fi
     fi
 fi
+
+# Make sure the appliance doesn't already have a restore underway
+if [ "$GHE_VERSION_MAJOR" -ge 2 ] && ghe-ssh "$GHE_HOSTNAME" -- "sudo grep -q restoring  $GHE_REMOTE_DATA_USER_DIR/common/ghe-restore-status 2>/dev/null"; then
+    echo "Error: $GHE_HOSTNAME already has a restore underway. Aborting." 1>&2
+    exit 1
+fi
+
+# Update remote restore state file and setup failure trap
+trap "update_restore_status failed" EXIT
+update_restore_status "restoring"
 
 # Make sure the GitHub appliance is in maintenance mode.
 if $instance_configured; then
--- a/bin/ghe-restore
+++ b/bin/ghe-restore
@@ -171,7 +171,7 @@ if [ "$GHE_VERSION_MAJOR" -ge 2 ] && ghe-ssh "$GHE_HOSTNAME" -- "sudo grep -q re
 fi
 
 # Update remote restore state file and setup failure trap
-trap "update_restore_status failed" EXIT
+trap "update_restore_status failed" EXIT SIGINT SIGTERM
 update_restore_status "restoring"
 
 # Make sure the GitHub appliance is in maintenance mode.
@@ -277,7 +277,7 @@ fi
 # Update the remote status to "complete". This has to happen before importing
 # ssh host keys because subsequent commands will fail due to the host key
 # changing otherwise.
-trap "" EXIT
+trap "" EXIT SIGINT SIGTERM
 update_restore_status "complete"
 
 # Log restore complete message in /var/log/syslog on remote instance
--- a/bin/ghe-restore
+++ b/bin/ghe-restore
@@ -171,7 +171,7 @@ if [ "$GHE_VERSION_MAJOR" -ge 2 ] && ghe-ssh "$GHE_HOSTNAME" -- "sudo grep -q re
 fi
 
 # Update remote restore state file and setup failure trap
-trap "update_restore_status failed" EXIT 2 15
+trap "update_restore_status failed" EXIT 1 2 3 15
 update_restore_status "restoring"
 
 # Make sure the GitHub appliance is in maintenance mode.
@@ -277,7 +277,7 @@ fi
 # Update the remote status to "complete". This has to happen before importing
 # ssh host keys because subsequent commands will fail due to the host key
 # changing otherwise.
-trap "" EXIT 2 15
+trap "" EXIT 1 2 3 15
 update_restore_status "complete"
 
 # Log restore complete message in /var/log/syslog on remote instance
--- a/bin/ghe-restore
+++ b/bin/ghe-restore
@@ -171,7 +171,7 @@ if [ "$GHE_VERSION_MAJOR" -ge 2 ] && ghe-ssh "$GHE_HOSTNAME" -- "sudo grep -q re
 fi
 
 # Update remote restore state file and setup failure trap
-trap "update_restore_status failed" EXIT 1 2 3 15
+trap "update_restore_status failed" EXIT HUP INT QUIT TERM
 update_restore_status "restoring"
 
 # Make sure the GitHub appliance is in maintenance mode.
@@ -277,7 +277,7 @@ fi
 # Update the remote status to "complete". This has to happen before importing
 # ssh host keys because subsequent commands will fail due to the host key
 # changing otherwise.
-trap "" EXIT 1 2 3 15
+trap "" EXIT HUP INT QUIT TERM
 update_restore_status "complete"
 
 # Log restore complete message in /var/log/syslog on remote instance
--- a/bin/ghe-restore
+++ b/bin/ghe-restore
@@ -171,7 +171,7 @@ if [ "$GHE_VERSION_MAJOR" -ge 2 ] && ghe-ssh "$GHE_HOSTNAME" -- "sudo grep -q re
 fi
 
 # Update remote restore state file and setup failure trap
-trap "update_restore_status failed" EXIT SIGINT SIGTERM
+trap "update_restore_status failed" EXIT 2 15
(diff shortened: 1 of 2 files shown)


bin/ghe-restore
```diff
@@ -271,7 +277,7 @@ fi
 # Update the remote status to "complete". This has to happen before importing
 # ssh host keys because subsequent commands will fail due to the host key
 # changing otherwise.
-trap "" EXIT
+trap "" EXIT 1 2 3 15
```
gnawhleinad: Could we use `SIGHUP`, `SIGINT`, `SIGQUIT`, and `SIGTERM` directly?

bin/ghe-restore
```diff
@@ -168,6 +164,16 @@ if [ "$GHE_VERSION_MAJOR" -le 1 ] && [ "$GHE_BACKUP_STRATEGY" = "rsync" ]; then
...
 fi
 
+# Make sure the appliance doesn't already have a restore underway
+if [ "$GHE_VERSION_MAJOR" -ge 2 ] && ghe-ssh "$GHE_HOSTNAME" -- "sudo grep -q restoring  $GHE_REMOTE_DATA_USER_DIR/common/ghe-restore-status 2>/dev/null"; then
+    echo "Error: $GHE_HOSTNAME already has a restore underway. Aborting." 1>&2
+    exit 1
+fi
+
+# Update remote restore state file and setup failure trap
+trap "update_restore_status failed" EXIT 1 2 3 15
```
gnawhleinad: Same as https://github.com/github/backup-utils/pull/176/files#r54185567.

bin/ghe-restore
snh: @gnawhleinad: I did originally try this but got the following error upon running the script on my development box:

    trap: SIGINT: bad trap

gnawhleinad: Ah, we're using `/bin/sh`. Does `HUP`, `INT`, `QUIT`, and `TERM` work?

snh: Sure does, my development system is using _dash_ which doesn't support using the SIG prefixes symbolic names.

rubiojr:
@snh so the issue here is not that we're not trapping signals correctly, the issue is that `ghe-restore-status` is left behind so future restore attempts won't work. This PR doesn't address that issue.

Does it make sense now? sorry if I wan't clear before.

>  In its current form this will prevent multiple restores in a cluster environment where the same host is specified as the restore host, but not where different hosts are specified that are part of the same cluster. For automated scenarios, which I believe is likely to be the most common scenario where multiple restores run at the same time, I imagine the same host is generally used.

The issue here is that the restore host could be the load balancer public IP, not a cluster node, so we may land in a different server next time, rendering this check ineffective. You may be able use `ghe-cluster-each` to check for the file if not found in the current node. This will add some additional roundtrips and may still be a bit racey, but probably ok.

snh:
> @snh so the issue here is not that we're not trapping signals correctly, the issue is that `ghe-restore-status` is left behind so future restore attempts won't work. This PR doesn't address that issue AFAICT.

@rubiojr I'm still a little confused sorry... Having `ghe-restore-status` left behind shouldn't be an issue as long as the trap fires and changes the contents of that file to _failed_. This PR only prevents a restore if `ghe-restore-status` contains _restoring_, the existence of the file isn't being used as the lock but rather the contents. The file needs to be left in place after `ghe-restore` exits so that the management console is aware if a restore has failed or is complete. In either of those states this PR won't prevent a new restore being fired off.

rubiojr:
> Having ghe-restore-status left behind shouldn't be an issue as long as the trap fires and changes the contents of that file to failed

Yup, you're right. `dash` [behaves differently to bash apparently](https://bugs.launchpad.net/ubuntu/+source/dash/+bug/608894) and that confused me. In `bash` `EXIT` is a pseudosignal that fires the command when the shell exits. So `HUP INT QUIT TERM` are not really required. They're required however when using dash, that is, `/bin/sh` in Ubuntu.

To clarify, the behaviour using `/bin/bash`:

```
+ ghe-ssh rubiojr-cluster.aws.octocluster.com:122 -- ghe-import-redis
+ true
+ echo 'Restoring Git repositories into cluster ...'
Restoring Git repositories into cluster ...
+ ghe-restore-repositories-dgit rubiojr-cluster.aws.octocluster.com:122
^C/home/rubiojr/backup-utils/share/github-backup-utils/ghe-restore-repositories-dgit: line 58: 70224 Killed                  ghe-ssh "$GHE_HOSTNAME" github-env ./bin/dgit-cluster-import-routes - < ssh_routes_out > ssh_routes_in
/home/rubiojr/backup-utils/share/github-backup-utils/ghe-restore-repositories-dgit: line 58: 70225 Killed                  ghe-ssh "$GHE_HOSTNAME" github-env ./bin/dgit-cluster-import-finalize - < ssh_finalize_out > ssh_finalize_in
/home/rubiojr/backup-utils/share/github-backup-utils/ghe-restore-repositories-dgit: line 58: 73972 Killed                  ghe-rsync -aHR --delete -e "ssh -q $opts -p $port -F $config_file -l $user" --rsync-path="sudo -u git rsync" "$GHE_DATA_DIR/$GHE_RESTORE_SNAPSHOT/repositories/./$network_path" "$route:$GHE_REMOTE_DATA_USER_DIR/repositories"
/home/rubiojr/backup-utils/share/github-backup-utils/ghe-restore-repositories-dgit: line 58: 73973 Killed                  ghe-rsync -aHR --delete -e "ssh -q $opts -p $port -F $config_file -l $user" --rsync-path="sudo -u git rsync" "$GHE_DATA_DIR/$GHE_RESTORE_SNAPSHOT/repositories/./$network_path" "$route:$GHE_REMOTE_DATA_USER_DIR/repositories"
/home/rubiojr/backup-utils/share/github-backup-utils/ghe-restore-repositories-dgit: line 58: 73975 Killed                  ghe-rsync -aHR --delete -e "ssh -q $opts -p $port -F $config_file -l $user" --rsync-path="sudo -u git rsync" "$GHE_DATA_DIR/$GHE_RESTORE_SNAPSHOT/repositories/./$network_path" "$route:$GHE_REMOTE_DATA_USER_DIR/repositories"
+ update_restore_status failed
+ '[' 2 -ge 2 ']'
+ echo failed
+ ghe-ssh rubiojr-cluster.aws.octocluster.com:122 -- 'sudo dd of='\''/data/user/common/ghe-restore-status'\'' 2>/dev/null'
```

When using `dash` and removing `HUP INT QUIT TERM` handlers:

```
Restoring Git repositories into cluster ...
+ ghe-restore-repositories-dgit rubiojr-cluster.aws.octocluster.com:122
^C/home/rubiojr/backup-utils/share/github-backup-utils/ghe-restore-repositories-dgit: line 58: 105398 Killed                  ghe-ssh "$GHE_HOSTNAME" github-env ./bin/dgit-cluster-import-routes - < ssh_routes_out > ssh_routes_in
/home/rubiojr/backup-utils/share/github-backup-utils/ghe-restore-repositories-dgit: line 58: 105399 Killed                  ghe-ssh "$GHE_HOSTNAME" github-env ./bin/dgit-cluster-import-finalize - < ssh_finalize_out > ssh_finalize_in
/home/rubiojr/backup-utils/share/github-backup-utils/ghe-restore-repositories-dgit: line 58: 106204 Killed                  ghe-rsync -aHR --delete -e "ssh -q $opts -p $port -F $config_file -l $user" --rsync-path="sudo -u git rsync" "$GHE_DATA_DIR/$GHE_RESTORE_SNAPSHOT/repositories/./$network_path" "$route:$GHE_REMOTE_DATA_USER_DIR/repositories"
/home/rubiojr/backup-utils/share/github-backup-utils/ghe-restore-repositories-dgit: line 58: 106205 Killed                  ghe-rsync -aHR --delete -e "ssh -q $opts -p $port -F $config_file -l $user" --rsync-path="sudo -u git rsync" "$GHE_DATA_DIR/$GHE_RESTORE_SNAPSHOT/repositories/./$network_path" "$route:$GHE_REMOTE_DATA_USER_DIR/repositories"
/home/rubiojr/backup-utils/share/github-backup-utils/ghe-restore-repositories-dgit: line 58: 106206 Killed                  ghe-rsync -aHR --delete -e "ssh -q $opts -p $port -F $config_file -l $user" --rsync-path="sudo -u git rsync" "$GHE_DATA_DIR/$GHE_RESTORE_SNAPSHOT/repositories/./$network_path" "$route:$GHE_REMOTE_DATA_USER_DIR/repositories"
```

Note the trap command wasn't fired here.

So, some of the backup-utils scripts are already using `/usr/bin/env bash` to avoid this kind of issues and things like https://github.com/github/backup-utils/issues/160, so I'd be in favour of using `#!/usr/bin/bash` in `ghe-restore`. Trapping `HUP INT QUIT TERM` should not be necessary at that point.

snh:
> Some of the backup-utils scripts are already using #!/usr/bin/env bash to avoid this kind of issues and things like #160, so I'd be in favour of using #!/usr/bin/env bash in ghe-restore. Trapping HUP INT QUIT TERM should not be necessary at that point.

:+1: Sounds like a good approach to me since bash is pretty universal.

For consistency it might be worth changing all `backup-utils` components so they use bash and adding it as a requirement to https://github.com/github/backup-utils#backup-host-requirements

rubiojr:
@snh yeah that's planned. Feel free to beat me to it :wink:

snh:
> The issue here is that the restore host could be the load balancer public IP, not a cluster node, so we may land in a different server next time, rendering this check ineffective. You may be able use ghe-cluster-each to check for the file if not found in the current node. This will add some additional roundtrips and may still be a bit racey, but probably good enough.

I'm wondering if a better approach would be to maintain `$GHE_REMOTE_DATA_USER_DIR/common/ghe-restore-status` on all cluster nodes? At the moment `ghe-store` only maintains it on the first server it lands on during a restore, so in theory if someone was to visit `http(s)://[hostname]/setup` during a restore, and hit a different node, it wouldn't be aware of the restore status, or even that a restore was underway?

snh:
Closing this PR for now, as it is unlikely that this is a common enough problem to warrant the extra code and complexity.

PR #177, opened 2016-03-01
Change to bash

snh:
Discussed in #176, this PR changes the shell used by _backup-utils_ from `#!/bin/sh` to `#!/usr/bin/env bash`.

Passes tests but requires extensive testing against a variety of GHE instances before merging.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,7 +45,7 @@ storage and must have network connectivity with the GitHub Enterprise appliance.
 ##### Backup host requirements
 
 Backup host software requirements are modest: Linux or other modern Unix
-operating system with [rsync][4] v2.6.4 or newer.
+operating system with bash and [rsync][4] v2.6.4 or newer.
 
 The backup host must be able to establish network connections outbound to the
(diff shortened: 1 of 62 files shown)


rubiojr:
Also fixes #160

lildude:
This also fixes https://github.com/github/backup-utils/issues/85

lildude:
One thought: we should probably document a minimum bash version. I know things like Solaris still ship with older versions of bash by default. This may not be necessary if we don't take advantage of any newer functionality though.

snh:
> One thought: we should probably document a minimum bash version.

:+1: I'll do some research and testing to see what baseline we should set.

rubiojr:
This is looking good to me.

Let's merge this for now and we'll figure out the bash version requirements later. @janorn and @andrewdotn already reported this would work for them and won't be a problem in a reasonably modern linux distribution with Bash 4.

snh:
Agreed. I suspect that we will have to go back quite a long way in to bash's past to find a minimum version.

Should I go ahead and :shipit: ?

gnawhleinad:
:shipit:!

PR #179, opened 2016-03-07
Use .sync_in_progress file during restores

snh:
This PR prevents Git GC operations and some other maintenance jobs from running while repositories are being restored using `ghe-restore-repositories-rsync`, `ghe-restore-repositories-dgit` and `ghe-restore-repositories-gist`.

This makes use of the existing `$GHE_REMOTE_DATA_USER_DIR/repositories/.sync_in_progress` file, which is already used by `ghe-backup-repositories-rsync` and `ghe-backup-repositories-cluster`, and largely copies the existing code from these.

Also waits for Git GC operations to finish before beginning restore, much like backups already do.

/cc @github/backup-utils


snh:
Updated to include cluster support.

snh:
Also waits for Git GC operations to finish before beginning restore, much like backups already do.

share/github-backup-utils/ghe-restore-repositories-dgit
```diff
@@ -55,15 +57,65 @@ echo "$config" > "$config_file"
...
+  for hostname in $hostnames; do
+    ghe-ssh -F $config_file "$hostname:$port" -- "sudo rm -f '$sync_in_progress_file'"
+  done
   rm -rf $config_file ssh_routes_in ssh_routes_out ssh_finalize_in ssh_finalize_out
 }
 
 trap 'cleanup' INT TERM EXIT
 
+wait_pids=""
+for hostname in $hostnames; do
```
rubiojr: I believe this will match all the servers in the cluster, not only git servers. Is that the intention?

Running something like:

```
ghe-cluster-each -r git -p
```

will return the list of git servers only. Same applies when cleaning up the `sync_in_progress_file`.

share/github-backup-utils/ghe-restore-repositories-dgit
```diff
@@ -55,15 +57,65 @@ echo "$config" > "$config_file"
...
+          # grep process itself.
+          if ps axo args | grep -E -e '^git( -.*)? nw-repac[k]( |$)' -e '^git( -.*)? g[c]( |$)' >/dev/null; then
+              sleep 1
+              sanity=\$(( sanity + 1 ))
+          else
+              exit 0
+          fi
+      done
+      exit 7
+  " | ssh -q $opts -p $port -F $config_file -l $user $hostname -- /bin/sh
```
rubiojr: We may want to use the `ghe-ssh` helper here, and `/bin/bash` (`/bin/sh` is Dash).

share/github-backup-utils/ghe-restore-repositories-gist
```diff
@@ -57,15 +59,65 @@ echo "$config" > "$config_file"
...
 
+wait_pids=""
+for hostname in $hostnames; do
+  # Touch the sync-in-progress file, disabling GC operations, and wait for all
+  # active GC processes to finish on the remote side.
+  echo "
+      set -e
+      sudo -u git touch '$sync_in_progress_file'
+      sanity=0
+      while [ \$sanity -lt $GHE_GIT_COOLDOWN_PERIOD ]; do
```
rubiojr: `for i in $(seq $GHE_GIT_COOLDOWN_PERIOD); do` may help to avoid using the `sanity` counter. Am I wrong?

share/github-backup-utils/ghe-restore-repositories-dgit
```diff
@@ -55,15 +57,65 @@ echo "$config" > "$config_file"
...
+      res=$?
+      if [ $res = 7 ]; then
+          echo "Error: Git GC processes remain after $GHE_GIT_COOLDOWN_PERIOD seconds on host: $wait_hostname. Aborting..." 1>&2
+      fi
+      exit $res
+  }
+  unset wait_hostname wait_pid
+done
+
+unset wait_pids
```
rubiojr: No need for `unset` I think. Both of them.

share/github-backup-utils/ghe-restore-repositories-dgit
rubiojr: @snh why are we not handling the `exit 7` error here?

share/github-backup-utils/ghe-restore-repositories-dgit
```diff
@@ -55,15 +57,65 @@ echo "$config" > "$config_file"
...
+      exit 7
+  " | ssh -q $opts -p $port -F $config_file -l $user $hostname -- /bin/sh
+  wait_pids="$wait_pids $hostname:$!"
+done
+
+# strip leading space
+wait_pids="$(echo -e "${wait_pids}" | sed -e 's/^[[:space:]]*//')"
+
+# wait for jobs one by one, bail on the first one
+for pid in $wait_pids; do
```
rubiojr: @snh I'm trying to understand why do we need this second loop with the `wait` here. Isn't it enough to handle the error condition like we do in `ghe-backup-repositories-rsync`?

share/github-backup-utils/ghe-restore-repositories-gist
rubiojr: This also applies to the other while loops we added to this PR. I know this is mostly a copy/pasta from other scripts but the `for` loop looks a bit cleaner to me unless I'm missing something.

rubiojr: We can probably add a helper a this point to avoid duplicating this everywhere. No need to refactor the existing scripts, I can take care of that.

snh:
@rubiojr Thanks for the detailed and valuable feedback. I very much appreciate it! ✨ 

I have just pushed a couple of new commits that refactor a number of areas as a result of @rubiojr's feedback.

As a lot of this code was copy and pastes from existing code in the repository backup scripts, there is some refactoring to be done to those areas too. This is best done as a seperate activity since this PR only deals with the restore side.

rubiojr:
@snh :metal:, I really like how you approached this. Let me :eyes: that one more time.

share/github-backup-utils/ghe-gc-disable
```diff
@@ -0,0 +1,48 @@
+#!/bin/sh
```
rubiojr: `#!/usr/bin/env bash` here?

share/github-backup-utils/ghe-gc-disable
```diff
@@ -0,0 +1,48 @@
...
+#/ Usage: ghe-gc-disable [<option>...] <host>
+#/ Helper to disable and drain GC operations on a GitHub Enterprise server.
+set -e
+
+# Bring in the backup configuration
+cd $(dirname "$0")/../..
```
rubiojr: The `cd` is already causing trouble in other scripts when installing `backup-utils` from the Debian package. Now that we're using `Bash`, can use use something like this instead to avoid altering `PWD`?

```
BASE_PATH="$( cd "$( dirname "${BASH_SOURCE[0]}" )/../.." && pwd )"
. share/github-backup-utils/ghe-backup-config
```

share/github-backup-utils/ghe-gc-disable
```diff
@@ -0,0 +1,48 @@
+#!/bin/sh
+#/ Usage: ghe-gc-disable [<option>...] <host>
+#/ Helper to disable and drain GC operations on a GitHub Enterprise server.
```
rubiojr: Even if it's for internal use, we may want to explain the options here:

```
#/ Usage: Usage: ghe-gc-disable [<option>...] <host>
#/
#/ Helper to disable and drain GC operations on a GitHub Enterprise server.
#/
#/ OPTIONS:
#/   -F <required_arg>                Blah blah
```

We still need to add a styleguide, but that'll be the preferred format probably.

share/github-backup-utils/ghe-restore-repositories-gist
```diff
@@ -61,11 +63,20 @@ cleanup() {
...
+    ghe-gc-enable -F $config_file $hostname:$port
+  done
   rm -rf $config_file ssh_routes_in ssh_routes_out ssh_finalize_in ssh_finalize_out
 }
 
 trap 'cleanup' INT TERM EXIT
 
+# Disable remote GC operations
+for hostname in $hostnames; do
+  ghe-gc-enable -F $config_file $hostname:$port
```
rubiojr: `ghe-gc-disable` here?

snh:
@rubiojr Thanks for the latest feedback. I've made a few changes as a result. Does the way I have re-jigged the working directory stuff look OK? My plan would be to do a new PR to update all the other scripts.

share/github-backup-utils/ghe-gc-disable
```diff
@@ -9,8 +9,8 @@
 set -e
 
 # Bring in the backup configuration
-cd $(dirname "$0")/../..
-. share/github-backup-utils/ghe-backup-config
+export GHE_BACKUP_ROOT="$( cd "$( dirname "${BASH_SOURCE[0]}" )/../.." && pwd )"
```
rubiojr: I would avoid the `export` here to prevent potential regression/behaviour change in a patch release.

share/github-backup-utils/ghe-gc-enable
```diff
@@ -9,8 +9,8 @@
 set -e
 
 # Bring in the backup configuration
-cd $(dirname "$0")/../..
-. share/github-backup-utils/ghe-backup-config
+export GHE_BACKUP_ROOT="$( cd "$( dirname "${BASH_SOURCE[0]}" )/../.." && pwd )"
```
rubiojr: I would avoid the `export` here to prevent potential regression/behaviour change in a patch release.

rubiojr:
> Does the way I have re-jigged the working directory stuff look OK?

Added some comments to avoid changing behaviour and potential regressions for now but the rest looks great to me.

> My plan would be to do a new PR to update all the other scripts.

That'd be great.

share/github-backup-utils/ghe-gc-disable
rubiojr: I mean avoid the `export` and don't use the `GHE_BACKUP_ROOT` variable name for now. That'd change `ghe-backup-config` behaviour I think.

snh: I did try and avoid changing ghe-backup-config behavior but ran in to a
issue with it requiring the working dir be changed to the root backup utils
dir before being called.

In share/github-backup-utils/ghe-gc-disable
<https://github.com/github/backup-utils/pull/179#discussion_r57606669>:

> @@ -9,8 +9,8 @@
>  set -e
>
>  # Bring in the backup configuration
> -cd $(dirname "$0")/../..
> -. share/github-backup-utils/ghe-backup-config
> +export GHE_BACKUP_ROOT="$( cd "$( dirname "${BASH_SOURCE[0]}" )/../.." && pwd )"

I mean avoid the export and don't use the GHE_BACKUP_ROOT variable name for
now. That'd change ghe-backup-config behaviour I think.

—
You are receiving this because you were mentioned.
Reply to this email directly or view it on GitHub
<https://github.com/github/backup-utils/pull/179/files/26feaa1f4392b21b825040e0c268ecf1cb0fd5da..3427dfe7df2867bd54250b37ee7c4ae7e150efef#r57606669>

rubiojr: @snh ah, gotcha. Then leaning towards reverting the change for now and keep using `pwd` for now and address this in a future PR.

rubiojr: That is, reverting https://github.com/github/backup-utils/pull/179/commits/3427dfe7df2867bd54250b37ee7c4ae7e150efef.

snh: @rubiojr Sounds like a plan. I will revert that commit.

rubiojr:
Thanks Steven, :shipit: when you want.

PR #182, opened 2016-03-19
Minimum version enforcement for cluster restores

### LINKED ISSUE #181: Block restoring 2.4 backups into a cluster
Restoring backups taken on a < 2.5 instance into a cluster is unsupported and will fail. This should be checked for and blocked.

snh:
Require that snapshots originated from an instance running GitHub Enterprise 2.5.0 or above when restoring to a cluster.

Fixes #181

/cc @github/backup-utils @cjs @rubiojr @electroniko


bin/ghe-restore
```diff
@@ -106,6 +106,17 @@ if ghe-ssh "$GHE_HOSTNAME" -- \
   restore_settings=false
 fi
 
+# Figure out if we're restoring into a cluster with an unsupported snapshot
+if $cluster; then
+  snapshot_instance_version=$(cat $GHE_RESTORE_SNAPSHOT_PATH/version)
+  if ! echo $snapshot_instance_version | \
+    grep -Eq "v2\.[5-9]|v2\.[1-9][0-9]|v[3-9]|v[1-9][0-9]"; then
+    echo "Error: Snapshot must be from GitHub Enterprise v2.5.0 and above to be restored"
```
gnawhleinad: v2.5.0 or above

gnawhleinad:
Could we add tests to [`test/test-ghe-restore.sh`](https://github.com/github/backup-utils/blob/master/test/test-ghe-restore.sh)?

snh:
> Could we add tests to `test/test-ghe-restore.sh`?

As we don't currently have any tests for cluster restores, it would be a big change to incorporate tests in to this PR, so probably best to do it as a seperate activity.

snh:
> Could we add tests to `test/test-ghe-restore.sh`?

I originally commented that we don't have tests for clusters in backup-utils, which wasn't entirely accurate. We don't have tests in this repo because these tests currently need a real cluster infrastructure to run against. That's not possible to setup for this repo, so we run those tests internally.

I will add a test for this PR to the existing internal tests.

snh:
Thanks to @gnawhleinad, this PR now includes tests. These tests run without requiring access to a real cluster infrastructure, so are able to be included in this repo. ✨

bin/ghe-restore
```diff
@@ -106,6 +106,17 @@ if ghe-ssh "$GHE_HOSTNAME" -- \
   restore_settings=false
 fi
 
+# Figure out if we're restoring into a cluster with an unsupported snapshot
+if $cluster; then
+  snapshot_instance_version=$(cat $GHE_RESTORE_SNAPSHOT_PATH/version)
+  if ! echo $snapshot_instance_version | \
+    grep -Eq "v2\.[5-9]|v2\.[1-9][0-9]|v[3-9]|v[1-9][0-9]"; then
```
rubiojr: Reminded me that we need to add a semver comparison helper at some point, somewhere. The ^^^ regex may not be future proof :trollface:, but who wants to live forever.

rubiojr:
🙇 👍

snh:
@rubiojr I just updated this branch so its up to date with master and updated a reference to `sh` to be to `bash` instead. Your welcome to merge this PR if you feel its ready for inclusion. 👍

PR #187, opened 2016-03-24
Add Bash styleguide

rubiojr:
/cc @github/backup-utils


snh:
Will be a very useful reference when working on `backup-utils` 👍

PR #189, opened 2016-03-30
Remove bashisms

snh:
Now that we have switched to bash exclusively, we can remove `bashisms` from the tests.

/cc @github/backup-utils


rubiojr:
@snh the test we removed looks good to me, but I wouldn't remove the `devscripts` package dependency. it has a bunch of scripts that will make the Debian packager life way easier, `dch` being one of them.

rubiojr:
🚢 :bow:

snh:
@rubiojr No worries, I've added these back in. Thanks for the heads up!

PR #190, opened 2016-03-30
Don't change working directory when loading in ghe-backup-config

snh:
Currently we change the working directory to the root backup-utils directory in each script. A number of scripts, such as `ghe-backup-config`, assume and therefore depend on this being the case.

This PR removes this, and instead sets `GHE_BACKUP_ROOT` to the root directory of backup-utils and uses this when referencing other scripts and files, instead of assuming the working directory is the root backup-utils directory.

Depends on #189 for the tests to pass.

/cc @github/backup-utils


share/github-backup-utils/ghe-backup-config
```diff
@@ -1,22 +1,23 @@
-#!/usr/bin/env bash
+#!/bin/sh
```
rubiojr: `#!/usr/bin/env bash` ?

bin/ghe-backup
```diff
@@ -8,8 +8,8 @@
 set -e
 
 # Bring in the backup configuration
-cd $(dirname "$0")/..
-. share/github-backup-utils/ghe-backup-config
+export GHE_BACKUP_ROOT="$( cd "$( dirname "${BASH_SOURCE[0]}" )/.." && pwd )"
```
rubiojr: We should export from `ghe-backup-config` only, and use `. $( dirname "${BASH_SOURCE[0]}" )/ghe-backup-config` to source the `ghe-backup-config` script (or whatever path is required if the script is not in `share/backup-utils`). That'll prevent globals pollution an indadvertedly changing that global value.

We need to pay attention to `.deb` installs here and make sure the changes work with that too. If `GHE_BACKUP_ROOT` points to `/usr` in that case, we should be fine. We'll need to make sure the config search path does not change though.

share/github-backup-utils/ghe-backup-store-version
```diff
@@ -4,12 +4,11 @@
...
 # Bring in the backup configuration
-cd $(dirname "$0")/../..
-. share/github-backup-utils/ghe-backup-config
+. $( dirname "${BASH_SOURCE[0]}" )/ghe-backup-config
 
-if [ -d .git ]; then
+if [ -d $GHE_BACKUP_ROOT/.git ]; then
   version_info="$BACKUP_UTILS_VERSION"
-  ref=$(git rev-parse HEAD || true)
+  ref=$(git --git-dir=$GHE_BACKUP_ROOT/.git rev-parse HEAD || true)
```
snh: I used `--git-dir` here instead of `-C` to ensure the greatest backwards compatibility with Git.

rubiojr:
@snh neat, looks good to me :sparkles:

rubiojr:
:shipit:

PR #192, opened 2016-04-12
ghe-host-check: handle ssh_exchange_identification error

rubiojr:
This may also happen when trying to connect to a GHE instance using port
22:

```
ssh_exchange_identification: Connection closed by remote host
```

Handling this error improves port 122 autodetection.

/cc @github/backup-utils


gnawhleinad:
:shipit:!

PR #193, opened 2016-04-12
Write backup-utils version to ghe host for diagnostics

mutle:
/cc @github/backup-utils


share/github-backup-utils/ghe-backup-store-version
```diff
@@ -0,0 +1,24 @@
...
+
+# Bring in the backup configuration
+cd $(dirname "$0")/../..
+. share/github-backup-utils/ghe-backup-config
+
+# Perform a host-check and establish GHE_REMOTE_XXX variables.
+ghe_remote_version_required "$host"
+
+# Grab the host
+host="$GHE_HOSTNAME"
```
rubiojr: @mutle doesn't look like we need this?

rubiojr:
Thanks @mutle, wanna squash this before merging so we only get your commits here?

share/github-backup-utils/ghe-backup-store-version
```diff
@@ -0,0 +1,24 @@
...
+#/ Usage: ghe-backup-store-version
+#/ Stores information about the used version of backup-utils on <host>
+set -e
+
+# Bring in the backup configuration
+cd $(dirname "$0")/../..
+. share/github-backup-utils/ghe-backup-config
+
+# Perform a host-check and establish GHE_REMOTE_XXX variables.
+ghe_remote_version_required "$host"
```
rubiojr: `$host` is going to be empty here, needs `$GHE_HOSTNAME`. No arguments will work just fine also, `ghe-host-check` (used by `ghe_remote_version_required`) uses `$GHE_HOSTNAME` if no arguments are provided.

rubiojr:
@mutle we may want to store the version when restoring (`ghe-restore`) also?

share/github-backup-utils/ghe-backup-store-version
rubiojr: Re-reading it I think we can just 🔥 `ghe_remote_version_required "$host"`, it's not required here.

PR #194, opened 2016-04-13
We want rsync to be verbose when using ghe-backup -v

rubiojr:
Useful with `ghe-backup -v` and also consistent with other scripts.

/cc @github/backup-utils


dbussink:
:+1:

PR #197, opened 2016-04-15
Check home directory for configuration file

snh:
Adds an extra supported location for the backup config.

backup-utils will attempt to load the backup configuration from the following locations, in this order:

1. `$GHE_BACKUP_CONFIG` (User configurable)
2. `$GHE_BACKUP_ROOT/backup.config` (Root directory of backup-utils install)
3. `$HOME/github-backup-utils/backup.config` (New)
4. `/etc/github-backup-utils/backup.config`

This is useful for situations where the user running backup-utils does not have write access to the install location or `/etc`, and does not wish to set an environment variable. Users of the Debian package are the main target audience for this PR.

/cc @github/backup-utils

### DIFF
--- a/share/github-backup-utils/ghe-backup-config
+++ b/share/github-backup-utils/ghe-backup-config
@@ -44,7 +44,7 @@ GHE_HOSTNAME_PRESERVE="$GHE_HOSTNAME"
 # first and then fall back to the backup-utils root, home directory and system.
 config_found=false
 for f in "$GHE_BACKUP_CONFIG" "$GHE_BACKUP_ROOT/backup.config" \
-  "$HOME/github-backup-utils/backup.config" "/etc/github-backup-utils/backup.config"; do
+  "$HOME/.github-backup-utils/backup.config" "/etc/github-backup-utils/backup.config"; do
     if [ -f "$f" ]; then
         GHE_BACKUP_CONFIG="$f"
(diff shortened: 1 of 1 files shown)


share/github-backup-utils/ghe-backup-config
```diff
@@ -46,7 +46,7 @@ GHE_HOSTNAME_PRESERVE="$GHE_HOSTNAME"
 # Source in the backup config file from the local working copy location first
 # and then falling back to the system location.
 config_found=false
-for f in "$GHE_BACKUP_CONFIG" "/etc/github-backup-utils/backup.config"; do
+for f in "$GHE_BACKUP_CONFIG" "$HOME/github-backup-utils/backup.config" "/etc/github-backup-utils/backup.config"; do
```
tjl2: In my experience, it'd be fairly standard behaviour for a config file in a user's home directory to override the settings in a system config file. Given that, would it be an idea to switch the order of the `$HOME/github-backup-utils/backup.config` and `/etc/github-backup-utils/backup.config` files in this loop?

rubiojr:
@snh using a hidden dir is perhaps a bit more common (`$HOME/.github-backup-utils/backup.config`). What do you think?

snh:
@rubiojr 👍

rubiojr:
:+1:

rubiojr:
:shipit:

PR #198, opened 2016-04-19
Revert "Don't change working directory when loading in ghe-backup-config"

rubiojr:
Reverts github/backup-utils#190

This may have changed behaviour and it's problematic when `GHE_DATA_DIR` is using a relative path like `data/2.5-backup`. Discovered it while trying to use the new fsck option:

```
rubiojr@enterprise2-dev:~/backup-utils$ ./bin/ghe-backup
Starting backup of 10.0.1.78 in snapshot 20160419T091140
Connect 10.0.1.78:122 OK (v2.6.0)
Backing up GitHub settings ...
Backing up SSH authorized keys ...
Backing up SSH host keys ...
Backing up MySQL database ...
Backing up Redis database ...
Backing up audit log ...
Backing up hookshot logs ...
Backing up Git repositories ...
Backing up GitHub Pages ...
Backing up asset attachments ...
Backing up storage data ...
Backing up hook deliveries ...
Backing up Elasticsearch indices ...
Running git fsck on repos...
Error: data/2.6-storage/20160419T091140 is not a valid snapshot.
Completed backup of 10.0.1.78:122 in snapshot 20160419T091140 at 09:13:13
Error: Snapshot incomplete. Some steps failed: fsck.
````

The snapshot directory is nested multiple times apparently:

```
rubiojr@enterprise2-dev:~/backup-utils$ ls data/2.6-storage/20160419T091140/
authorized-keys.json  data  incomplete  mysql.sql.gz  redis.rdb  ssh-host-keys.tar  strategy  version

rubiojr@enterprise2-dev:~/backup-utils$ ls data/2.6-storage/20160419T091140/data/2.6-storage/20160419T091140/
alambic_assets  data           enterprise.ghl  manage-password+  repositories   storage
audit-log       elasticsearch  hookshot        pages             settings.json

rubiojr@enterprise2-dev:~/backup-utils$ ls data/2.6-storage/20160419T091140/data/2.6-storage/20160419T091140/data/2.6-storage/
```

Setting `GHE_DATA_DIR=data/test` in the config file should be enough to reproduce it.

@snh: reverting for now but wanna take care of this?

PR #199, opened 2016-04-20
Don't change working directory when loading in ghe-backup-config 

snh:
2nd attempt at #190, which was reverted by #198 due to a bug when using relative paths as the data directory.

`share/github-backup-utils/ghe-backup-config` now canonicalizes the data directory path, basing any relative paths on the backup-utils root:

```
GHE_DATA_DIR="$( cd "$GHE_BACKUP_ROOT" && readlink -m "$GHE_DATA_DIR" )"
```


rubiojr:
LGTM.

@snh do you think we can add a test to prevent future regressions? Double checking that `GHE_DATA_DIR` layout is correct when using a relative path maybe?

snh:
@rubiojr Sounds like a good idea. I'll need to have a think about the best way to incorporate this in to the tests...

snh:
Updated to include relative path test.

I originally used `realpath` to construct a relative path, but discovered that a number of distributions include the non-GNU version, which lacks the `--relative-to=` option. As a result, the final implementation is a bit more crude, but should remain highly portable.

/cc @rubiojr

test/test-ghe-backup.sh
```diff
@@ -214,6 +213,76 @@ begin_test "ghe-backup subsequent snapshot"
...
+    # Change working directory to the root directory
+    cd $ROOTDIR
+
+    # run it
+    GHE_DATA_DIR=$(echo $GHE_DATA_DIR | sed 's|'$ROOTDIR'/||') ghe-backup
+
+    # check that current symlink points to new snapshot
+    this_snapshot=$(ls -ld "$GHE_DATA_DIR/current" | sed 's/.* -> //')
+    [ "$previous_snapshot" != "$this_snapshot" ]
+
```
rubiojr: Nit but rather than setting and comparing `previous_snapshot` and `this_snapshot`, I'd set `GHE_SNAPSHOT_TIMESTAMP`, making sure it's there after the backup. That way we don't need all the sed magic with the symlinks:

```
cd $ROOTDIR
export GHE_SNAPSHOT_TIMESTAMP=123-relative
export GHE_DATA_DIR=$(echo $GHE_DATA_DIR | sed 's|'$ROOTDIR'/||')
ghe-backup
```

test/test-ghe-backup.sh
```diff
@@ -214,6 +213,76 @@ begin_test "ghe-backup subsequent snapshot"
...
+    if [ "$GHE_VERSION_MAJOR" -ge 2 ]; then
+        # verify all hookshot user data was transferred
+        diff -ru "$GHE_REMOTE_DATA_USER_DIR/hookshot" "$GHE_DATA_DIR/current/hookshot"
+
+        # verify all git hooks data was transferred
+        diff -ru "$GHE_REMOTE_DATA_USER_DIR/git-hooks" "$GHE_DATA_DIR/current/git-hooks"
+
+        # verify all alambic assets user data was transferred
+        diff -ru "$GHE_REMOTE_DATA_USER_DIR/alambic_assets" "$GHE_DATA_DIR/current/alambic_assets"
+    fi
```
rubiojr: The `GHE_REMOTE_DATA_USER_DIR` test are probably not required, since we're already testing restores somewhere else?

rubiojr:
Added a couple of minor/style thingies but other than that, LGTM.

test/test-ghe-backup.sh
rubiojr: Nevermind, it's probably a great idea to not test directory presence only.

snh: 👍  I decided to leave these in here to make sure there weren't any fringe cases where the relative directory resulted in some data being written to the wrong location. I've gone back and added a couple of extra just to be safe too.

test/test-ghe-backup.sh
```diff
@@ -214,6 +213,86 @@ begin_test "ghe-backup subsequent snapshot"
...
+    sleep 1
+
+    # generate a timestamp
+    export GHE_SNAPSHOT_TIMESTAMP="relative-$(date +"%Y%m%dT%H%M%S")"
+
+    # change working directory to the root directory
+    cd $ROOTDIR
+
+    # run it
+    GHE_DATA_DIR=$(echo $GHE_DATA_DIR | sed 's|'$ROOTDIR'/||') ghe-backup
```
snh: My reasoning for only passing the relative path through to `ghe-backup` and not the rest of the test is so that the full absolute path is used for the tests.

rubiojr: I thin the subshell will protect us from that (we don't run `ghe-backup or `ghe-restore` again in this test AFAICT). But yeah, works for me too. The export was only to try to make it a bit more readable, but that's very subjective.

rubiojr:
👍  :shipit:

PR #200, opened 2016-04-26
Don't use --no-dangling with old git versions

rubiojr:
`--[no-]dangling` was added to Git 1.7.10.

/cc @github/backup-utils

### DIFF
--- a/share/github-backup-utils/ghe-backup-fsck
+++ b/share/github-backup-utils/ghe-backup-fsck
@@ -21,6 +21,12 @@ repos=0
 errors=0
 log=$(mktemp)
 t_start=$(date +%s)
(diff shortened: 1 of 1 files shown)


gnawhleinad:
LGTM!

rubiojr:
Doesn't work very well if git man pages are not available, so needs better detection. Trying to avoid parsing git version though.

rubiojr:
Although a properly packaged git will include them, so changed my mind. The downside is that we'll end up not using `--no-dangling` even if it's available, so it's not that bad.

rubiojr:
Ended up using `-h` instead of `--help` as suggested by @snh, so `git` doesn't need the man pages installed.

PR #204, opened 2016-05-09
clustering: fix git-hooks restore

rubiojr:
Wrong variable name and re-creating the wrong remote directory.

/cc @github/backup-utils


rubiojr:
Only affects restores to a GitHub Enterprise 2.6 cluster (using the new [Pre-receive hooks](https://help.github.com/enterprise/2.6/admin/guides/developer-workflow/creating-a-pre-receive-hook-environment) feature) if (and only if) the data being restored is from a 2.6 VM or cluster.

lildude:
LGTM and makes sense too. :+1:

PR #207, opened 2016-05-16
Support for systems lacking readlink -m

### LINKED ISSUE #205: The backup utils are broken on OS X
This [change](https://github.com/github/backup-utils/commit/f22e4a5696b2233738746b63f89ff8f1a78be425) introduced an error on OS X, which doesn't support the `-m` option for `readlink`.

snh:
Fixes #205, by falling back to a simplistic approach where `readlink -m` isn't available.

I investigated a number of other options, but our need to handle a path that may not yet exist caused most of them to be unsuitable.

/cc @github/backup-utils

### DIFF
--- a/share/github-backup-utils/ghe-backup-config
+++ b/share/github-backup-utils/ghe-backup-config
@@ -74,9 +74,12 @@ if [ -z "$GHE_HOSTNAME" ]; then
     exit 2
 fi
 
-# Canonicalize the data directory path, basing any relative paths on the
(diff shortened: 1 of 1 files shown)


rubiojr:
Should work, yeah 👍 

@snh also having a look at adding [Travis/OSX support](https://github.com/github/backup-utils/pull/208). Needs some work but doesn't look that bad: https://travis-ci.org/github/backup-utils/jobs/130562478

PR #209, opened 2016-05-16
Log git-hooks restores for cluster

rubiojr:
Minor, cosmetic issue but we should also log git-hooks restores for clusters.

/cc @github/backup-utils


gnawhleinad:
:shipit:!

PR #219, opened 2016-05-26
Benchmarking restores

rubiojr:
Simple benchmarking that logs the time it takes to restore data. Looks something like:

```
$ cat data/26-snap/current/benchmarks/benchmark.1464253606.log
ghe-import-mysql took: 7s
ghe-import-redis took: 43s
ghe-restore-repositories-dgit took: 11s
ghe-restore-alambic-cluster-ng took: 6s
ghe-restore-git-hooks-cluster took: 2s
ghe-restore-es-audit-log took: 0s
```

The log is timestamped and stored in the benchmarks directory, along with the data that was restored.

/cc @github/backup-utils


share/github-backup-utils/bm.sh
```diff
@@ -0,0 +1,51 @@
...
+# bm.sh: benchmarking Bash code blocks
+#
+# Example:
+#   bm_start "wget request"
+#   wget --quiet https://www.google.com
+#   bm_end "wget request"
+#
+# Sample output:
+#   $ bash test.sh
+#   'wget request' took: 2s
```
gnawhleinad: Should we drop the `:` so it reads

> ```
> wget request took 2s
> ```

rubiojr: :+1:

PR #223, opened 2016-06-06
Ghe restore variable option position

aknarts:
previously:
./ghe-restore -s 20130201T150109 10.0.0.1:122
now possible:
./ghe-restore 10.0.0.1:122 -s 20130201T150109

### DIFF
--- a/bin/ghe-restore
+++ b/bin/ghe-restore
@@ -28,6 +28,9 @@ set -e
 restore_settings=false
 force=false
 while true; do
+    if [ -z "$1" ]; then
+        break;
+    fi
     case "$1" in
         -f|--force)
             force=true
@@ -46,13 +49,12 @@ while true; do
             exit 1
             ;;
         *)
-            break
+            # Grab the host arg
+            GHE_HOSTNAME="${1:-$GHE_RESTORE_HOST}"
+            shift
             ;;
     esac
 done
-
-# Grab the host arg
-GHE_HOSTNAME="${1:-$GHE_RESTORE_HOST}"
 
 # Hostname without any port suffix
 hostname=$(echo "$GHE_HOSTNAME" | cut -f 1 -d :)
--- a/bin/ghe-restore
+++ b/bin/ghe-restore
@@ -50,11 +50,14 @@ while true; do
             ;;
         *)
             # Grab the host arg
-            GHE_HOSTNAME="${1:-$GHE_RESTORE_HOST}"
+            GHE_HOSTNAME="$1"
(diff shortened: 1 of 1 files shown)


aknarts:
I am pretty sure that the tests for ghe-restore fail every time.
For example when I look into them on line 97:
```
if ! ghe-restore -v -f > "$TRASHDIR/restore-out" 2>&1; then
```
ghe-restore does not work with parameter -v (it is evaluated as an invalid argument).

As for the change it is something we ran into when doing a test restore. It restored wrong snapshot (frankly we missed the confirmation where it said the snapshot that it is going to use). This does prevent this and grabs arguments even if they are used after the IP is specified.

rubiojr:
Thanks @aknarts.

> ghe-restore does not work with parameter -v (it is evaluated as an invalid argument).

So `-v` is handled from https://github.com/github/backup-utils/blob/master/share/github-backup-utils/ghe-backup-config#L28-L33. I'd say that it's confusing that we're doing it there, but `ghe-restore -v` should work as expected in any case. Also not having problems running `make test` locally, bear in mind that your change may have changed the argument parsing behaviour, breaking things.

> As for the change it is something we ran into when doing a test restore. It restored wrong snapshot (frankly we missed the confirmation where it said the snapshot that it is going to use). This does prevent this and grabs arguments even if they are used after the IP is specified.

I think we can improve argument handling here but we shouldn't change the way this works now:

* `ghe-restore --help` is well documented, the arguments are positional. Accepting arguments in random positions is doable but would complicate things in Bash unnecessarily, IMO. It's never been requested before so it's not a popular request either.
* There's a confirmation question, so you have a second chance to make sure you're restoring the correct snapshot.

We can make sure that no other argument is given after the host IP to prevent the issue you had from happening again. That'd be simpler and safer.

/cc @github/backup-utils for thoughts.

aknarts:
I see, the tests do not work correctly for me here, but that is quite possibly on my side.

I do agree with the solution that it should fail or throw out warning if there is another argument after the IP. 

The solution I was proposing did allow for some strange commandline syntax that did not help with the readability and usage. (like ./ghe-restore -c 10.0.0.1:122 -s 20130201T150109 ... which would make one think that IP is the argument for -c)

I will close this merge request as it is not the best solution for the problem.

rubiojr:
> I see, the tests do not work correctly for me here, but that is quite possibly on my side.

Out of curiosity, where are you running them? Linux, OSX, something else? could be a xplatform bug as well.

PR #227, opened 2016-06-14
restore host keys for cluster environment as well

### LINKED ISSUE #226: host keys are not restored under ghe cluster environment
@gnawhleinad pointed out that it doesn't restore host keys in cluster environment: [ghe-restore#L314-L317](https://github.com/github/backup-utils/blob/f7a401e30d8db3a20e4211482a3928c8f3476555/bin/ghe-restore#L314-L317).

If the host keys under `/data/user/common` are used by babeld process only, there is **no** difference between a cluster environment and a single appliance. since all `web-server` roles on frontend nodes must serve the same host key for all users.

shawnzhu:
fixes #226

### DIFF
--- a/bin/ghe-restore
+++ b/bin/ghe-restore
@@ -287,6 +287,10 @@ if [ "$GHE_VERSION_MAJOR" -ge 2 ]; then
         ghe-ssh "$GHE_HOSTNAME" -- /bin/sh
 fi
 
+# restore host keys at last step before applying config changes
+echo "Restoring SSH host keys ..."
+ghe-ssh "$GHE_HOSTNAME" -- 'ghe-import-ssh-host-keys' < "$GHE_RESTORE_SNAPSHOT_PATH/ssh-host-keys.tar" 1>&3
+
 # When restoring to a host that has already been configured, kick off a
 # config run to perform data migrations.
 if $cluster; then
@@ -310,11 +314,6 @@ update_restore_status "complete"
 
 # Log restore complete message in /var/log/syslog on remote instance
 ghe_remote_logger "Completed restore from $(hostname) / snapshot ${GHE_RESTORE_SNAPSHOT}."
-
-if ! $cluster; then
-  echo "Restoring SSH host keys ..."
-  ghe-ssh "$GHE_HOSTNAME" -- 'ghe-import-ssh-host-keys' < "$GHE_RESTORE_SNAPSHOT_PATH/ssh-host-keys.tar" 1>&3
-fi
 
 echo "Completed restore of $GHE_HOSTNAME from snapshot $GHE_RESTORE_SNAPSHOT"
 


rubiojr:
@shawnzhu thanks, we're looking into this.

The problem currently is that the server side script `ghe-import-ssh-host-keys` uses the same keys for the host and babeld, so we'll also need server side patches.

dbussink:
We've further discussed this, but it will be necessary to make some bigger changes. For clustering setups, we want the key to be restored that is used for Git over SSH traffic, but not share the same key across all cluster nodes for administrative SSH access. With change both keys are overwritten (for Git traffic and SSH administrative access), so it won't only restore the keys used for Git traffic. 

This is going to require bigger changes on the GitHub Enterprise side and will likely mean that this pull request is not the correct implementation to achieve this. Going to close this therefore for now until we have a better solution in place.

rubiojr:
> @rubiojr what about not using ghe-import-ssh-host-keys but just extracts ssh host keys on /data/user/common only in ghe-restore? which is just the first half of ghe-import-ssh-host-keys.

@shawnzhu yeah, that'd be part of the fix. Unfortunately there are other scripts that use the keys available in `/data/user/common` while booting (or setting up a new cluster node) that will copy the keys and put them in `/etc/ssh`, so the patch is going to be a bit more complicated.

This will definitely be fixed in the next feature release, but I'll have a deeper look this week and see if I can come up with a reasonable patch for a patch release or a safe workaround to be used in the meantime.

shawnzhu:
@rubiojr my immediate request is just using existing host key files for Git over SSH traffic when moving to cluster environment, so that users (especially those automated jobs triggered by SCM change from github) won't need to update their `~/.known_hosts` file.

rubiojr:
@shawnzhu: we had a closer look at this issue and this is likely going to be the right fix for backup-utils.

To give you a little bit of context, clusters (should) use a few different SSH host keys:
  * Some keys for the cluster node (in `/etc/ssh`).
  * Different keys for Git over SSH (babeld), that should be shared among all the cluster nodes (in `/data/user/common/ssh_host*`).
  
In single VM mode (regular installs, not clusters), the same key is used for both `sshd` and git over ssh but in a cluster, each cluster node should have it's own set of SSH host keys and share the keys used for Git over SSH. Merging this would mean that the key would be restored to `/data/user/common` and then `ghe-cluster-config-apply` would also copy those keys to `/etc/ssh` in every node.
This is what we were trying to prevent, since we would end up with all the cluster nodes having the same SSH host key (in addition to the same Git over SSH key).

Unfortunately, there's a bug in the cluster initialization code that is already replicating the same ssh host keys to all the cluster nodes, so what we were trying to prevent is already happening. You can currently check your ssh host keys using something like:

```
# SSH host keys
ghe-cluster-each -- sudo md5sum /etc/ssh/ssh_host* | sort -k2

# Git over SSH keys
ghe-cluster-each -- sudo md5sum /data/user/common/ssh_host* | sort -k2
```

I hope that helps you to understand the issue a bit better. We're currently working on a server side fix for 2.7 but in the meantime, we are going to merge this since we definitely want the Git over SSH keys to be restored. That should also fix the issues you're currently having.

Do you wan't to re-open the PR again so we can merge it?

Thanks!

shawnzhu:
@rubiojr Thanks for the in depth information!

I've checked the cluster host keys by these two commands and it matches what you described.

It would be great if you can re-open and merge this PR so that I don't need to do it manually.

Expecting for the coming v2.7 release!

rubiojr:
@shawnzhu can't do that, because you may have deleted the branch from your fork (I think). You could submit a new PR (preferred) or I could fix it myself, whatever works for you.

shawnzhu:
@rubiojr Thanks. I've created #228.

PR #232, opened 2016-06-21
Clusters: speedup repositories restore

rubiojr:
We now rsync once per cluster node available instead of rsyncing each
Git repository individually.

Some simple benchmarks restoring a snapshot with 1183 repositories (13 GiB):

* Using backup-utils 2.6.1

```
real    20m34.923s
user    5m1.888s
sys 2m39.983s

```

* Using the new implementation

```
real    9m0.368s
user    2m46.912s
sys 1m18.746s
```

The old implementation is able to restore ~1 repo/s so restoring backup
snapshots with a large number of repositories and a fast network will benefit
the most from this.

Here's the time it takes to restore 8K repositories (~800 MiB), all of them
very similar in size (100K, with a single README file added):

* Using backup-utils 2.6.1

```
real    111m45.370s
user    7m54.829s
sys 6m38.247s
```

* Using the new implementation

```
real    6m20.087s
user    0m16.509s
sys 1m42.616s
```

In clusters with more than 3 Git server nodes, backup-utils 2.6.1 also
restores the repositories to all the Git server nodes available. Only three
copies of a Git repository are necessary so this patch also fixes that,
speeding things up and optimizing disk usage.

/cc @github/backup-utils

### DIFF
--- a/bin/ghe-restore
+++ b/bin/ghe-restore
@@ -217,7 +217,12 @@ bm_end "ghe-import-redis"
 
 if $cluster; then
   echo "Restoring Git repositories into cluster ..."
-  ghe-restore-repositories-dgit "$GHE_HOSTNAME" 1>&3
+  if ghe-ssh "$GHE_HOSTNAME" test -f /data/github/current/script/dgit-cluster-restore-routes; then
+    ghe_verbose "* Using ghe-restore-repositories-dgit-ng to restore"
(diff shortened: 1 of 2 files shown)


rubiojr:
Forgot to mention that this needs GitHub Enterprise 2.6.4, otherwise it'll automatically use the old (slower) restore code.

PR #234, opened 2016-06-23
Help and Verbose output for commands in bin/

brntbeer:
I noticed while working on another branch that everything in `./bin/` was missing a `-v` or `-h`. This fixes it!

cc @github/backup-utils or putting this on whatever radar you may have!

### DIFF
--- a/bin/ghe-backup
+++ b/bin/ghe-backup
@@ -7,6 +7,28 @@
 #/ transferred.
 set -e
 
+# If -h or --help are passed into the command prompt, spit out
+# the help documents
+usage() {
+  grep '^#/' < $0 | cut -c4-
+  exit 2
+}
+
+# Depending on the variables passed in, either spit out the help, be more
+# verbose, or just continue
+while [ $# -gt 0 ]; do
+  case "$1" in
+    -h|--help)
+      usage
+      exit 2
+      ;;
+    -v|--verbose)
+      set -x
+      ;;
+  esac
+  shift
+done
+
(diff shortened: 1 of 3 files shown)


rubiojr:
Thanks @brntbeer, this PR would be a nice addition with some changes:

* `--verbose` (or `-v` to be precise) is already handled by https://github.com/github/backup-utils/blob/master/share/github-backup-utils/ghe-backup-config#L29 and it has a different meaning. It means "more verbose output" from the tools being used like `rsync` and such.
The way you implemented verbose means Bash tracing. Also verbose but a different thing that can be accomplished using something like `bash -x ./bin/ghe-restore` for example. A nice touch would be to have an additional `--trace` flag (instead of the `-v|--verbose` you added), so we keep the existing semantics but make Bash tracing a bit more discoverable.

* Having multiple scripts that parse options and shift arguments is confusing and problematic, so the best way to handle this would be to move the argument parsing from `ghe-backup-config` to `ghe-restore` and/or `ghe-backup` IMO.

Overhauling the way we parse arguments to be a bit cleaner and follow our [styleguide](https://github.com/github/backup-utils/blob/master/STYLEGUIDE.md) was in my mind, so greatly appreciated.

pakokoto:
thanks

brntbeer:
@rubiojr sorry for the delay here. I'm going to close this down and possibly implement something along the lines of `--trace` as you mapped out at some point in the future.

thanks for the guidance!

PR #240, opened 2016-07-11
Update software requirements in README.md

### LINKED ISSUE #239: Document that Git is a requirement for the backup host
Not all Linux distros include Git as a package that is installed by default, so this should be explicitly listed as a requirement in https://github.com/github/backup-utils#backup-host-requirements.

snh:
- Adds Git to requirements
- Adds link to Bash homepage
- Fixes #239 

/cc @github/backup-utils @djdefi


michaeltwofish:
:shipit:

PR #242, opened 2016-07-25
Remove restoring cluster.conf on restore

dbussink:
This breaks restoring in a differently layed out cluster which is explicitly supported.

cc @rubiojr @github/backup-utils


rubiojr:
👍

PR #238, opened 2016-07-08
Benchmark backups

snh:
Same as #219, but for backups and includes per-node stats for cluster restores.

Standard backup looks something like this:

```
ghe-backup-store-version took 0s
ghe-backup-settings took 1s
ghe-export-authorized-keys took 0s
ghe-export-ssh-host-keys took 0s
ghe-export-mysql took 1s
ghe-backup-redis took 1s
ghe-backup-es-audit-log took 2s
ghe-backup-es-hookshot took 0s
ghe-backup-repositories-rsync took 1s
ghe-backup-userdata - pages took 1s
ghe-backup-pages-rsync took 1s
ghe-backup-userdata - alambic_assets took 0s
ghe-backup-userdata - storage took 0s
ghe-backup-userdata - hookshot took 0s
ghe-backup-userdata - git-hooks/repos took 0s
ghe-backup-es-rsync took 1s
```

Cluster backup looks something like this:

```
ghe-backup-store-version took 0s
ghe-backup-settings took 3s
ghe-export-authorized-keys took 0s
ghe-export-ssh-host-keys took 0s
ghe-export-mysql took 7s
ghe-backup-redis-cluster took 2s
ghe-backup-es-audit-log took 1s
ghe-backup-es-hookshot took 1s
ghe-backup-repositories-cluster - data-0 took 5s
ghe-backup-repositories-cluster - data-1 took 5s
ghe-backup-repositories-cluster - data-2 took 4s
ghe-backup-repositories-cluster took 17s
ghe-backup-pages-cluster - data-0 took 1s
ghe-backup-pages-cluster - data-1 took 1s
ghe-backup-pages-cluster - data-2 took 1s
ghe-backup-pages-cluster took 4s
ghe-backup-alambic-cluster - data-0 took 1s
ghe-backup-alambic-cluster - data-1 took 1s
ghe-backup-alambic-cluster - data-2 took 1s
ghe-backup-alambic-cluster took 3s
ghe-backup-git-hooks-cluster took 3s
```

The logs are timestamped and stored in the benchmarks directory, along with the data that was backed up.

/cc @github/backup-utils @rubiojr

PR #243, opened 2016-07-27
Bump version: 2.6.4

rubiojr:
Patch release fixing a regression and instrumenting backups.

Changes the cluster restore behaviour a bit since it doesn't restore the
cluster configuration (cluster.conf) any more.
Backup restores to the same cluster and are not affected.
Backup restores from a single VM to a cluster (i.e. migrating from a single
GitHub Enterprise VM to a cluster using backup-utils) are also unaffected.

* Instrument/benchmark backups #238
* Cluster: remove restoring cluster.conf on restore #242

/cc @github/backup-utils


gnawhleinad:
:shipit:!

PR #244, opened 2016-07-28
Prevent restoring a cluster snapshot to a standalone GHE appliance

jatoben:
This case isn't supported currently, and will fail with a somewhat confusing error message. I've added a check and early exit to make things a bit more clear.

/cc @github/backup-utils


rubiojr:
Nice @jatoben, thanks.

rubiojr:
Going to include this in the upcoming 2.6.4 release also.

PR #248, opened 2016-08-24
Add backup version note

pravipati:
Version numbers are important when restoring a backup—`ghe-restore` only supports restoring from a backup running an older version if that version is no more than two releases behind the version of GitHub Enterprise running on the backup host. 

/cc @dbussink


README.md
```diff
@@ -75,6 +75,9 @@ supported. We strongly recommend upgrading to the latest release if you're
 running a version prior to 11.10.342. Visit [enterprise.github.com][5] to
 download the most recent GitHub Enterprise version.
 
+Note: You can only restore from snapshot that's at most two releases ahead of the version of GitHub Enterprise running on your backup host. For example, to restore from a snapshot running GitHub Enterprise 2.7, your backup host must be on GitHub Enterprise 2.6 or 2.5. You can't restore from 2.4 to 2.7, because that's three releases behind.
```
brntbeer: > You can only restore from snapshot

You can only restore from **a** snapshot.

djdefi: This reads a little confusing to me, as the "backup host" is not running GitHub Enterprise (in most cases 😄 ). 

Some edits I would suggest:

>Note: You can only restore a snapshot that's at most two feature releases behind the restore target's version of GitHub Enterprise. For example, to restore a snapshot of GitHub Enterprise 2.4, the target GitHub Enterprise appliance must be running GitHub Enterprise 2.5.x or 2.6.x. You can't restore a snapshot from 2.4 to 2.7, because that's three releases ahead.

rubiojr: :+1:

rubiojr:
@pravipati looks like a good addition to me. With @djdefi's and @brntbeer's suggestions reads a bit better.

djdefi:
This has been sitting a while, so I went ahead and pushed the new wording to the branch. Can we get a final review / approval?

cc: @rubiojr @brntbeer @pravipati

PR #265, opened 2016-11-14
Use Travis container infrastructure

jatoben:
This switches to [Travis's container infrastructure](https://docs.travis-ci.com/user/migrating-from-legacy/), which typically results in faster builds.

/cc @github/backup-utils @snh


rubiojr:
Build times are roughly the same in our case but it may (or may not) help Travis 😉.

PR #276, opened 2016-12-12
Remove redundant hookshot elasticsearch index backups and restores

snh:
When restoring to a non-cluster environment, we don't need to individually restore Elasticsearch based hookshot indices, as these will be included in the overall Elasticsearch restore.

As some housekeeping, also updated the backup process so that `ghe-backup-es-hookshot` is only called during backups of 2.x appliances.

/cc @github/backup-utils


rubiojr:
@snh makes sense to me, but I'd like to get some 👀 from @mutle who worked on that in the past.

PR #278, opened 2016-12-30
Use Trusty for Travis builds

jatoben:
In #265 we briefly tried Travis's Trusty-based build environment, but [had to revert it](https://github.com/github/backup-utils/pull/265/commits/8fcdedc1efdcd90effed0ca2f0cea1675feae546) because the `devscripts` package wasn't yet available. It is now (cc https://github.com/travis-ci/apt-package-whitelist/issues/3653), so this PR makes the Trusty switch.

/cc @github/backup-utils @snh

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,7 @@ matrix:
       script: make test
     - os: linux
       dist: trusty
-      sudo: false
+      sudo: true
       addons:
         apt:
           packages:
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,5 +11,7 @@ matrix:
       sudo: false
       addons:
         apt:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -7,8 +7,11 @@ matrix:
         - brew install gnu-tar
       script: make test
     - os: linux
-      sudo: false
+      dist: trusty
+      sudo: true
```
jatoben: We don't need `sudo`, but [adding it triggers use of a GCE VM](https://docs.travis-ci.com/user/trusty-ci-environment#Overview) rather than a Docker container, and the container took >10 minutes longer to finish a build.

PR #279, opened 2017-01-05
Update some bash bangpaths to use /usr/bin/env

smkelly:
There are two files that have their bang paths set to #!/bin/bash. This is not consistent with the rest of the project and also results in these tools failing to work on FreeBSD and other platforms that don't default to bash use.


rubiojr:
Awesome @smkelly, thanks.

Issue #163, opened 2016-02-18
Custom CA certificates aren't included in backup-utils


cjs:
backup-utils does not backup or restore over custom root CA certificates added with `ghe-ssl-ca-certificate-install`.

PR #281, opened 2017-02-01
Backup and restore custom CA certificates

snh:
Using enhancements added in GitHub Enterprise 2.8.0, adds support for the backup and restore of custom CA certificates.

These certificates are only restored if an appliance is unconfigured, or the restore is performed with the `-c` flag.

Requires GitHub Enterprise 2.8.0 or later.

Fixes https://github.com/github/backup-utils/issues/163

### DIFF
--- a/share/github-backup-utils/ghe-backup-settings
+++ b/share/github-backup-utils/ghe-backup-settings
@@ -45,7 +45,7 @@ if [ "$GHE_VERSION_MAJOR" -ge 2 ]; then
     fi
 fi
 
-if ghe-ssh "$host" -- "which ghe-export-ssl-ca-certificates"; then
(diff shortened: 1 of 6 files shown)


share/github-backup-utils/ghe-backup-settings
```diff
@@ -45,6 +45,11 @@ if [ "$GHE_VERSION_MAJOR" -ge 2 ]; then
     fi
 fi
 
+if ghe-ssh "$host" -- "which ghe-export-ssl-ca-certificates"; then
```
lildude: I've not tested this, but isn't this going to produce unnecessary output in the backup output?  We might want to direct all output to /dev/null to be sure.

rubiojr: Good catch, that will write the path of that script when found.

PR #287, opened 2017-02-27
Do not remove the working copy's .git dir

rubiojr:
Safer to work on a throw-away dir inside TRASHDIR.

Fixes some test issues we've seen introduced by ebca53e441dbbfef6c63d7ea6f93c971cf68a431.

/cc @github/backup-utils @jatoben

### DIFF
--- a/test/test-ghe-backup.sh
+++ b/test/test-ghe-backup.sh
@@ -463,7 +463,7 @@ begin_test "ghe-backup stores version when not run from a clone"
   # Make sure this doesn't exist
   rm -f "$GHE_REMOTE_DATA_USER_DIR/common/backup-utils-version"
 
-  tmpdir=(mktemp -d $TRASHDIR/foo.XXXXXX)
+  tmpdir=$(mktemp -d $TRASHDIR/foo.XXXXXX)
   git clone $ROOTDIR $tmpdir/backup-utils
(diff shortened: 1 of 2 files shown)


test/test-ghe-backup.sh
```diff
@@ -463,9 +463,12 @@ begin_test "ghe-backup stores version when not run from a clone"
   # Make sure this doesn't exist
   rm -f "$GHE_REMOTE_DATA_USER_DIR/common/backup-utils-version"
 
-  mv "$ROOTDIR/.git" "$ROOTDIR/.gittmp"
-  ghe-backup
-  mv "$ROOTDIR/.gittmp" "$ROOTDIR/.git"
+  tmpdir=$TRASHDIR/ghe-backup-stores-version-$RANDOM
```
sbryant: I think `mktemp -d` might work here

rubiojr: @sbryant indeed, but I want it to be inside `TRASHDIR`, so it's automatically garbage collected. We could use additional args but macOS compatibility will make it harder. The randomly generated name should be totally safe and doesn't have compat issues.

rubiojr: Ended up using @sbryant's `mktemp -d` suggestion, that works with BSD and Linux `mktemp`.

PR #289, opened 2017-02-27
Suppress dd output noise

rubiojr:
```
Starting restore of rubiojr5.foobar.com:122 from snapshot 20170227T112756
rubiojr5-app-1: 0+1 records in
rubiojr5-app-1: 0+1 records out
rubiojr5-app-1: 10 bytes (10 B) copied, 7.1914e-05 s, 139 kB/s
Error: rubiojr5.foobar.com:122 must be put in maintenance mode before restoring. Aborting.
rubiojr5-app-1: 0+1 records in
rubiojr5-app-1: 0+1 records out
rubiojr5-app-1: 7 bytes (7 B) copied, 8.4164e-05 s, 83.2 kB/s
```

dd doesn't support `status=none` in GHE <= 2.5 (Precise) I think, so we may need a wrapper here, or just wait for 2.5 to be EOLd.

/cc @github/backup-utils


snh:
I have confirmed that the updated `sponge` based method functions as expected when restoring to a single instance. Have not verified to a cluster as yet.

.travis.yml
```diff
@@ -21,4 +24,5 @@ matrix:
           packages:
             - devscripts
             - debhelper
+            - moreutils
```
rubiojr: Do we need moreutils/sponge installed on a Travis host?

snh: Yes, it turns out this isn't on any of the targets we test on. See https://travis-ci.org/github/backup-utils/builds/206119031 for build failures before I added this to the travis hosts.

rubiojr: Of course, forgot about mocking/stubs... :bow:

PR #294, opened 2017-03-17
Solaris 11 fixes and new configuration documentation

janorn:
I have made two changes to make backup-utils work on Solaris 11.
First change ps command= to ps comm= which is posix compliant.

Second I change the PATH if backup-utils is run on Solaris 11.
We add a path to GNU utils first and /usr/sbin due to unlink residing in /usr/sbin.

### DIFF
--- a/backup.config-example
+++ b/backup.config-example
@@ -43,3 +43,9 @@ GHE_NUM_SNAPSHOTS=10
 #
 # WARNING: do not enable this, only useful for debugging/development
 #GHE_BACKUP_FSCK=no
+
+# In some operatingsystems various tools such as gnu are located elsewhere
+# Modify you path here.
+#
+# Set this path on Solaris 11
+#PATH="/usr/gnu/bin:$PATH:/usr/sbin"
--- a/bin/ghe-backup
+++ b/bin/ghe-backup
@@ -65,7 +65,7 @@ if [ -f ../in-progress ]; then
   progress=$(cat ../in-progress)
   snapshot=$(echo "$progress" | cut -d ' ' -f 1)
   pid=$(echo "$progress" | cut -d ' ' -f 2)
-  if ! ps -p $pid -o command= | grep ghe-backup; then
+  if ! ps -p $pid -o comm= | grep ghe-backup; then
     # We can safely remove in-progress, ghe-prune-snapshots
     # will clean up the failed backup.
     unlink ../in-progress


share/github-backup-utils/ghe-backup-config
```diff
@@ -22,6 +22,10 @@ BACKUP_UTILS_VERSION="$(cat $GHE_BACKUP_ROOT/share/github-backup-utils/version)"
 
 # Add the bin and share/github-backup-utils dirs to PATH
 PATH="$GHE_BACKUP_ROOT/bin:$GHE_BACKUP_ROOT/share/github-backup-utils:$PATH"
+# In Solaris 11 reassure that gnu utils are used and add /usr/sbin for unlink
+if [ "SunOS 5.11" = "$(uname -sr)" ]; then
+  PATH="/usr/gnu/bin:$PATH:/usr/sbin"
+fi
```
lildude: Having this here sets a precedent for similar behaviour and operations on other operating systems which we don't want.  This could turn into a messy collection of OS-specific `if`s.

I think this is better suited to live within your own copy of `backup.config` as this is the location best suited for customer/site specific customisations; you might have the GNU utils at `/usr/gnu/bin` on your Solaris 11 system, but it doesn't mean everyone does (the various GNU pkg are optional IIRC).

How about adding a commented out `PATH=` entry and generic comment to the end of the `backup.config-example` instead?

janorn: I need to state that /usr/gnu/bin is where solaris installs gnu tools provided by Oracle. It is nothing optional about this path. I will investigate the backup.config-example.

lildude: I agree 100%.  The point I was trying to make is you don't need to install any of the GNU pkgs to get a fully functional Solaris 11 system.  If you so desire, you can install the GNU utils from source into `/usr/local/bin` or the old traditional location for third party software of `/opt` instead. In these cases `/usr/gnu/bin` wouldn't be a valid path.

lildude: ... and it's because of :point_up: that I think it would be best to define the PATH specific to your system in the `backup.config` file.

janorn:
I have now moved the PATH config as suggest to the example file. Please note that /usr/gnu/bin is where Oracle installs all provided gnu tools to separate them from the builtin tools. This is for Solaris 11 not Solaris 10. Many gnu tools are provided directly from Oracle.

backup.config-example
```diff
@@ -43,3 +43,9 @@ GHE_NUM_SNAPSHOTS=10
 #
 # WARNING: do not enable this, only useful for debugging/development
 #GHE_BACKUP_FSCK=no
+
+# In some operatingsystems various tools such as gnu are located elsewhere
+# Modify you path here.
+#
+# Set this path on Solaris 11
+#PATH="/usr/gnu/bin:$PATH:/usr/sbin"
```
lildude: I think it's best to make this as generic as possible without any references to specific operating systems.  I know Solaris puts the GNU utils here (I used to work for Sun and then Oracle on Solaris 😄) but we shouldn't mention this in the configuration example as it's not generic.

How about modifying this to:

```
# The PATH for the backup utilities to use. The default PATH used is that of the 
# executing user which may not always be customizable or contain all the utilities
# required by the backup utilities.
#
# PATH=""
```

This leaves things very generic and open to customisation that best suites the operating system or environment.

janorn: So where should information for future users be placed so they don't have todo all investigation over again... 

I worked for Sun+Oracle 10+ years.

lildude: As we document in the README.md, we rely on the comments in the configuration file to explain the purpose of each settings so I think this is a good place and generic enough for most sysadmins to find and understand, regardless of operating system.  I think it's only a matter of getting the wording correct to keep things generic whilst at the same time conveying that this is what needs to be modified, regardless of OS, if commands are not in the path of the executing user.

> I worked for Sun+Oracle 10+ years.

So did I. 👋 fellow Sun alumnus.

bin/ghe-backup
```diff
@@ -65,7 +65,7 @@ if [ -f ../in-progress ]; then
   progress=$(cat ../in-progress)
   snapshot=$(echo "$progress" | cut -d ' ' -f 1)
   pid=$(echo "$progress" | cut -d ' ' -f 2)
-  if ! ps -p $pid -o command= | grep ghe-backup; then
+  if ! ps -p $pid -o comm= | grep ghe-backup; then
```
rubiojr: @janorn this change will break backup-utils in Linux hosts unfortunately, it's not equivalent:

```
rubiojr@enterprise2-dev5-cp1-prd:~/backup-utils$ ps -p 103090 -o comm=
bash
rubiojr@enterprise2-dev5-cp1-prd:~/backup-utils$ ps -p 103090 -o command=
bash ./bin/ghe-backup
```

rubiojr: What about `ps -p 103090 -o args`? according to https://docs.oracle.com/cd/E26502_01/html/E29030/ps-1.html, that may work on Solaris too?

janorn: @rubiojr seems to be the same on solaris. Never tested this on a running backup. args works on solaris.

```
ghebup@backup:~$ cat data/in-progress 
20170411T140748 7198
ghebup@backup:~$ ps -p 7198 -o comm
COMMAND
bash
ghebup@backup:~$ ps -p 7198 -o args
COMMAND
bash /home/ghebup/backup-utils/bin/ghe-backup -v
```

rubiojr: @janorn thanks. Feel free to update the branch so we can merge it.

lildude:
Whoops, this PR seems to have stagnated. Do you have the cycles to address the current suggestions and conflicts so we can progress this PR @janorn?

If not, no probs. I can spin off a new PR with all the changes.

lildude:
Closing in favour of https://github.com/github/backup-utils/pull/436

PR #304, opened 2017-05-16
Change the condition in *ghe-host-check* when access through port 22 fails 

pnsk:
The current condition:

```
            if echo "$output" | grep -i "port 22: connection refused\|port 22: no route to host\|ssh_exchange_identification: Connection closed by remote host\|Connection timed out during banner exchange" >/dev/null; then
                exec "$(basename $0)" "$hostname:122"
            fi
```

can't catch the simple "Connection time out" error like:

```
ssh: connect to host xxx.xxx.xxx.xxx port 22: Connection timed out
```

So, I changed the condition a bit.

### DIFF
--- a/bin/ghe-host-check
+++ b/bin/ghe-host-check
@@ -36,7 +36,7 @@ set -e
 if [ $rc -ne 0 ]; then
     case $rc in
         255)
-            if echo "$output" | grep -i "port 22: connection refused\|port 22: no route to host\|ssh_exchange_identification: Connection closed by remote host\|Connection timed out" >/dev/null; then
+            if echo "$output" | grep -i "port 22: connection refused\|port 22: no route to host\|ssh_exchange_identification: Connection closed by remote host\|Connection timed out during banner exchange\|port 22: Connection timed out" >/dev/null; then
                 exec "$(basename $0)" "$hostname:122"
             fi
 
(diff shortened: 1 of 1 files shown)


snh:
This can be reproduced by adding a `-j DROP` `iptables` rule on your backup host:

```
sudo iptables -I OUTPUT -p tcp -d [ghe-ip] --dport 22 -j DROP
```

bin/ghe-host-check
```diff
@@ -36,7 +36,7 @@ set -e
 if [ $rc -ne 0 ]; then
     case $rc in
         255)
-            if echo "$output" | grep -i "port 22: connection refused\|port 22: no route to host\|ssh_exchange_identification: Connection closed by remote host\|Connection timed out during banner exchange" >/dev/null; then
+            if echo "$output" | grep -i "port 22: connection refused\|port 22: no route to host\|ssh_exchange_identification: Connection closed by remote host\|Connection timed out" >/dev/null; then
```
snh: To minimise the chances of this hiding issues with `122`, can we leave the `banner exchange` match as is, and add a new one for `port 22: Connection timed out`, so that it only matches on `22`?

Without this, it can get stuck in a loop if `22` and `122` both time out.

pnsk: > Without this, it can get stuck in a loop if 22 and 122 both time out.

Ah yes, you're right. It'll loop.

PR #306, opened 2017-05-17
Retry loop for redis-cli BGSAVE

buckelij:
`ghe-backup-redis` dumps can fail and hang indefinitely if BGSAVE is called while Redis "background rewriting" is happening. In these cases BGSAVE will report an error, but still exit 0. 

This PR adds a `for` loop around the `redis-cli BGSAVE` to retry if BGSAVE reports an `ERR`. I also put the `LASTSAVE` check in a `for` loop that times out after 60 minutes. 

I'm not certain how long the `for` loop `sleep`s should be, but I think 60 minutes is more than enough.

### DIFF
--- a/share/github-backup-utils/ghe-backup-redis
+++ b/share/github-backup-utils/ghe-backup-redis
@@ -22,13 +22,13 @@ sudo=
 ghe-ssh "$GHE_HOSTNAME" /bin/bash <<EOF
     set -e
     timestamp=\$(redis-cli LASTSAVE)
-    for i in $(seq 10); do
+    for i in \$(seq 10); do
       if ! redis-cli BGSAVE | grep -q ERR; then
         break
       fi
       sleep 15
     done
-    for n in $(seq 240); do
+    for n in \$(seq 240); do
       if [ "\$(redis-cli LASTSAVE)" != "\$timestamp" ]; then
         break
       fi
--- a/share/github-backup-utils/ghe-backup-redis
+++ b/share/github-backup-utils/ghe-backup-redis
@@ -19,12 +19,25 @@ ghe_remote_version_required "$GHE_HOSTNAME"
 # Force a redis BGSAVE, and wait for it to complete.
 sudo=
 [ "$GHE_VERSION_MAJOR" -ge 2 ] && sudo="sudo"
-ghe-ssh "$GHE_HOSTNAME" /bin/sh <<EOF
+ghe-ssh "$GHE_HOSTNAME" /bin/bash <<EOF
     set -e
     timestamp=\$(redis-cli LASTSAVE)
-    redis-cli BGSAVE 1>/dev/null
-    while [ \$(redis-cli LASTSAVE) -eq \$timestamp ]; do
-        sleep 1
+    for ((n=0; n<10; n++)); do
+      if redis-cli BGSAVE | grep -q ERR; then
+        sleep 15
+      else
+        break
+      fi
+    done
+    for ((n=0; n<240; n++)); do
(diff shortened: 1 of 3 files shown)


buckelij:
So, `debuild` from `package-deb` was running the tests as well. That seems redundant, and was causing the checks to timeout. I added `DEB_BUILD_OPTIONS=nocheck` to `debuild` in f3a351c.

share/github-backup-utils/ghe-backup-redis
```diff
@@ -19,12 +19,25 @@ ghe_remote_version_required "$GHE_HOSTNAME"
...
+    for ((n=0; n<240; n++)); do
+      if [ \$(redis-cli LASTSAVE) -eq \$timestamp ]; then
+          sleep 15
+      else
+        break
+      fi
+      if [ \$n -eq 239 ]; then
+        exit 1
+      fi
     done
```
rubiojr: This for loop style is a bit more compact and preferred.

```bash
for n in $(seq 240); do
  if [ "\$(redis-cli LASTSAVE)" != "\$timestamp" ]; then
    break
  fi
done
[ "\$(redis-cli LASTSAVE) != "\$timestamp" ] # exits 1 if bgsave didn't work
```

share/github-backup-utils/ghe-backup-redis
```diff
@@ -19,12 +19,25 @@ ghe_remote_version_required "$GHE_HOSTNAME"
...
-    redis-cli BGSAVE 1>/dev/null
-    while [ \$(redis-cli LASTSAVE) -eq \$timestamp ]; do
-        sleep 1
+    for ((n=0; n<10; n++)); do
+      if redis-cli BGSAVE | grep -q ERR; then
+        sleep 15
+      else
+        break
+      fi
+    done
```
rubiojr: This for loop style is preferred:

```bash
for i in $(seq 10); do
  if ! redis-cli BGSAVE | grep -q ERR; then
    break
  fi
  sleep 15
done
```

buckelij:
@rubiojr made the style changes and re-enabled the test. Something must have been going on with CI earlier -- the travis-ci checks were passing but the backup-utils test was timing out during the deb package build.

share/github-backup-utils/ghe-backup-redis
```diff
@@ -19,13 +19,23 @@ ghe_remote_version_required "$GHE_HOSTNAME"
...
-    while [ \$(redis-cli LASTSAVE) -eq \$timestamp ]; do
-        sleep 1
+    for i in \$(seq 10); do
+      if ! redis-cli BGSAVE | grep -q ERR; then
+        break
+      fi
+      sleep 15
     done
+    for n in \$(seq 240); do
+      sleep 1
```
buckelij: I added a sleep here and took a second off the 2nd sleep. The fake `redis-cli` used in the test environment increments every second, but this was running through too fast and adding an extra 15 seconds to every backup-utils run. That was adding up and causing the tests to time out.

buckelij:
Ready for review again @rubiojr

share/github-backup-utils/ghe-backup-redis
rubiojr: @buckelij any reason why we can't just replace `240` with `3600` to simplify things?

rubiojr: And leave only the `sleep 1`, that is.

buckelij: yeah, that's fine. Fixed in 5627d6d.

rubiojr:
:bow:

PR #314, opened 2017-07-31
explain how to extract the download

brntbeer:
It could probably be argued that someone dealing and managing a set of backups on a server should know how to extract the file, but i don't see any loss (only benefit) when explaining what command to run to extract the file.

cc @github/backup-utils for your 👀 (for some reason i can't assign the team as reviewers ☹️ )

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,8 +80,11 @@ Note: You can restore a snapshot that's at most two feature releases behind the 
 
 ### Getting started
 
- 1. [Download the latest release version][release] and extract *or* clone the
-    repository using Git:
+ 1. [Download the latest release version][release] and extract the repository using `tar`:
+ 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -80,8 +80,11 @@ Note: You can restore a snapshot that's at most two feature releases behind the
 
 ### Getting started
 
- 1. [Download the latest release version][release] and extract *or* clone the
-    repository using Git:
+ 1. [Download the latest release version][release] and extract the repository using `tar`:
+ 
+    `tar -xvf /path/to/github-backup-utils-vMAJOR.MINOR.PATCH.tar.gz`
```
rubiojr: I think we're missing -z here (compressed).

brntbeer:
@rubiojr updated. Let me know if the ordering is off on those flags (cant remember if it really matters).

PR #315, opened 2017-07-31
Update travis.yml distributions and packages

rubiojr:
debuild now complains that fakeroot is missing for some reason.

See https://travis-ci.org/github/backup-utils/jobs/259352688

/cc @github/backup-utils @brntbeer


rubiojr:
Thanks @snh 🙇

PR #317, opened 2017-08-01
Removed obsoleted code

sridharavinash:
This code is obsoleted now that we get the consolidated routes for all the repos including _Archived Repos_ from one call.

/cc @rubiojr @snh


snh:
@github/backup-utils We'll hold off on merging this until we have confirmed that all Clusters are using the new method.

PR #321, opened 2017-08-08
SSH multiplexing support

snh:
Multiplexing will cut down on the latency involved in opening and closing the many SSH connections used by the backup utilities, and more importantly ensures that connections to a cluster always land on the same node, which will facilitate PRs such as https://github.com/github/backup-utils/pull/318.

A seperate mux is created per username, hostname and IP combination, resulting in a single multiplexed connection to each host. Each control master will persist in the background until it has been idle for 10 minutes, which allows the backup utilities to cope with brief periods when there are no active SSH connections during a backup or restore while local processing is happening.

A short sha of sorts is used as part of the name of mux control socket to work around socket path length limitations. These sockets are placed in `GHE_DATA_DIR` as hidden files.

Multiplexing can be disabled by setting the environment variable `GHE_DISABLE_SSH_MUX`.

I also took the opportunity to refactor the code used to generate the SSH configurations.

/cc @github/backup-utils @github/enterprise-on-prem

### DIFF
--- a/share/github-backup-utils/ghe-backup-alambic-cluster
+++ b/share/github-backup-utils/ghe-backup-alambic-cluster
@@ -24,10 +24,6 @@ fi
 # Perform a host-check and establish GHE_REMOTE_XXX variables.
 ghe_remote_version_required "$host"
 
-# Generate SSH config for forwarding
-
-config=""
-
 # Split host:port into parts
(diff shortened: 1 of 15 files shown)


rubiojr:
:zap:

PR #324, opened 2017-08-13
Prefer ssh port specified at command line

snh:
This PR fixes a bug with the new SSH multiplexing implementation that would see the wrong port used for the control socket name if a port was specified using `-p` and not as part of the hostname:

```
$ ./share/github-backup-utils/ghe-ssh test-host:122 -- hostname
test-host
$ ./share/github-backup-utils/ghe-ssh -p 122 test-host -- hostname
```

This bug is triggered in a number of the `rsync` calls used during the backup of a standalone appliance.

/cc @github/backup-utils @jatoben for review


frconil:
LGTM @snh !

PR #327, opened 2017-08-25
Add note about file system case sensitivity

### LINKED ISSUE #241: Docs: Backup host filesystem should be case-sensitive
Backing up to case-insensitive filesystems such as _HFS+ (Case-insensitive)_ can be troublesome. 

There can be cases where a filename and directory name have the same name but different cases, which causes rsync to fail.

djdefi:
Closes #241

Recommends a case sensitive file system. Open to suggestions on the wording or if more detail is wanted.

cc: @github/backup-utils

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,6 +61,8 @@ and growth over time.
 The backup utilities use [hard links][12] to store data efficiently, so the backup
 snapshots must be written to a filesystem with support for hard links.
 
+Using a [case sensitive][26] file system is strongly recommended to avoid conflicts.
+
 ##### GitHub Enterprise version requirements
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -61,6 +61,8 @@ and growth over time.
 The backup utilities use [hard links][12] to store data efficiently, so the backup
 snapshots must be written to a filesystem with support for hard links.
 
+Using a [case sensitive][26] file system is strongly recommended to avoid conflicts.
```
kathodos: `s/[26]/[16]/`

djdefi:
Thanks @kathodos -- fixed now

PR #330, opened 2017-09-07
move host check so that ES dump scripts can run standalone

terrorobe:
we need to give `ghe_remote_version_required` a chance to rewrite `$GHE_HOSTNAME` to include port 122 before using it in later `ghe-ssh` calls, otherwise either script fails with

```
++ ghe-ssh ghefusion.michaelrenner.at 'curl -s "localhost:9201/_cat/indices/audit_log*?h=index"'
```

### DIFF
--- a/share/github-backup-utils/ghe-backup-es-audit-log
+++ b/share/github-backup-utils/ghe-backup-es-audit-log
@@ -11,11 +11,11 @@ set -e
 
 bm_start "$(basename $0)"
 
-# Set up remote host and root elastic backup directory based on config
-host="$GHE_HOSTNAME"
-
 # Perform a host-check and establish GHE_REMOTE_XXX variables.
 ghe_remote_version_required "$host"
+
+# Set up remote host and root elastic backup directory based on config
+host="$GHE_HOSTNAME"
 
 # Make sure root backup dir exists if this is the first run
 mkdir -p "$GHE_SNAPSHOT_DIR/audit-log"
--- a/share/github-backup-utils/ghe-backup-es-hookshot
+++ b/share/github-backup-utils/ghe-backup-es-hookshot
@@ -11,11 +11,11 @@ set -e
 
 bm_start "$(basename $0)"
 
-# Set up remote host and root elastic backup directory based on config
-host="$GHE_HOSTNAME"
-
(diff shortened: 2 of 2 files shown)


lildude:
Nudging this as I'm in a "new year spring cleaning mood" 😄

terrorobe:
> `$host` needs to be set before `ghe_remote_version_required` runs, otherwise `$host` as passed in to `ghe_remote_version_required` will be `null`.

Entertainingly enough this is what makes this work 😄 

---

From a contract perspective, `ghe_remote_version_required` can take input, but doesn't need it. The function will ensure that `$GHE_HOSTNAME` is set to a connectable string after it returns. If no hostname is passed in, it calls `ghe-host-check` with an empty string which in turn will use the hostname found in the config.

We're a bit inconsistent with how we're calling `ghe_remote_version_required`.

`ghe-backup` calls it without argument, `ghe-restore` passes in `$GHE_HOSTNAME`.

17 scripts call it with `$GHE_HOSTNAME`.

11 scripts call it with `$host` with a few differences in how things are handled:

Some of them set `$host` to `$GHE_HOSTNAME` first, e.g. [`ghe-backup-es-audit-log`](https://github.com/github/backup-utils/blob/master/share/github-backup-utils/ghe-backup-es-audit-log),  some of them call it with an empty `$host` variable (e.g. [`ghe-backup-git-hooks-cluster`](https://github.com/github/backup-utils/blob/master/share/github-backup-utils/ghe-backup-git-hooks-cluster)) and some of them do something different altogether (e.g. [`ghe-backup-repositories-cluster-ng`](https://github.com/github/backup-utils/blob/master/share/github-backup-utils/ghe-backup-repositories-cluster-ng)).

---

I'm unsure what the best way forward is. My gut feeling would be to changing all scripts to call `ghe_remote_version_required` without an argument unless it's possible to pass a hostname in through an command line argument. That's how `ghe-backup` & `ghe-restore` are handling things at the moment.

Furthermore, ensure that all subsequent script-internal hostname variables are only set after `ghe_remote_version_required` returns. These two changes should make all scripts runnable without the `ghe-backup` or `ghe-restore` wrapper.

Thoughts?

PR #335, opened 2017-09-15
Use git to generate short name for SSH multiplex control path 

### LINKED ISSUE #334: Backup fails if the exact command `sha256sum` is not found
Backup fails if the exact command `sha256sum` is not found. 
```Starting backup of GitHub.int.****.com in snapshot 20170914T105717
/Users/****/repos/ghe-backup-utils/share/github-backup-utils/ghe-ssh: line 61: sha256sum: command not found
```
This didn't happen yesterday, and appears to be caused by the changes since yesterday.
I have a ssh config to use the standard admin port. 
I run the same commands every day:
```~/repos/ghe-backup-utils $ cd ~/repos/ghe-backup-utils/; git pull origin stable ; bin/ghe-backup ; du -chs /Users/****/repos/ghe-data-backup/data/*```
The results for this latest backup show directory's content size of zero bytes.

MacOS does not include the `sha256sum` command by default, and none is found using `brew search sha256sum`. On the Mac, the equivalent command is `shasum -a 256 [FILE]`.

snh:
Switches over to ~`md5sum`~ `git` for the generation of a shorted name, as ~`md5sum`~ `git` is more widely universally available than `sha256sum`, which is not available by default on Mac OS.

Fixes #334.

### DIFF
--- a/share/github-backup-utils/ghe-ssh
+++ b/share/github-backup-utils/ghe-ssh
@@ -58,7 +58,7 @@ if echo "$*" | grep "[|;]" >/dev/null || [ $(echo "$*" | wc -l) -gt 1 ]; then
 fi
 
 if [ -z "$GHE_DISABLE_SSH_MUX" ]; then
-  controlpath="$TMPDIR/.ghe-sshmux-$(echo -n "$user@$host:$port" | sha256sum | cut -c 1-8)"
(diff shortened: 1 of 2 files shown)


snh:
I've updated this to use `git` instead, as I didn't realise that `md5sum` also has a different name on Mac OS (`md5`).

PR #341, opened 2017-09-28
Ignore another error from ps

mgriffin:
Similar to #260, this is another instance where `ps` uses the `-o` flag. This PR follows the same fix as before and redirects any error to `/dev/null`.

/cc @github/backup-utils

### DIFF
--- a/bin/ghe-backup
+++ b/bin/ghe-backup
@@ -69,7 +69,7 @@ if [ -f ../in-progress ]; then
   progress=$(cat ../in-progress)
   snapshot=$(echo "$progress" | cut -d ' ' -f 1)
   pid=$(echo "$progress" | cut -d ' ' -f 2)
-  if ! ps -p $pid -o command= 2>/dev/null | grep ghe-backup; then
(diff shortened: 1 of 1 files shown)


bin/ghe-backup
```diff
@@ -69,7 +69,7 @@ if [ -f ../in-progress ]; then
   progress=$(cat ../in-progress)
   snapshot=$(echo "$progress" | cut -d ' ' -f 1)
   pid=$(echo "$progress" | cut -d ' ' -f 2)
-  if ! ps -p $pid -o command= | grep ghe-backup; then
+  if ! ps -p $pid -o command= 2>/dev/null | grep ghe-backup; then
```
snh: It'd be good if we could still provide protection against multiple simultaneous backups to those who lack `-o` support.

As an alternative, could we drop the use of `-o` altogether?

```sh
if ! ps -p $pid | grep ghe-backup; then
```

mgriffin:
@snh I've removed the `-o` flag, what do you think now?

Issue #345, opened 2017-10-22
github-backup-utils 2.11.x exits too quickly when the destination is mounted on nfs


williamh:
Ever since we upgraded our github enterprise instance and upgraded github-backup-utils to 2.11.x, we have an issue with ghe-backup exiting before everything is written to our backup snapshot, which is mounted on nfs.

Basically the following procedure fails:

```
mount /path/to/backup
# this is an nfs mount
time ghe-backup
umount /path/to/backup
# This fails because the nfs mount is busy, and I have verified that nothing other than the ssh process 
# from ghe-backup is holding the mount.
```

Is this a regression in github-backup-utils, or should we  work around it on our side?

Thanks,

William

lildude:
This sounds like the SSH multiplexing support that was introduced in https://github.com/github/backup-utils/pull/321 which used the `$GHE_DATA_DIR` for the temporary multiplex files. This was then moved to `$TMP` in https://github.com/github/backup-utils/pull/323 and is included in the latest release of backup-utils.

Assuming you're not setting `$TMP` to a location on your NFS mount, updating to the latest backup-utils release should solve this problem for you. If you can't update just yet, you should find disabling SSH multiplexing by setting `GHE_DISABLE_SSH_MUX=1` in your environment will solve the issue too.

If the problem persists, then it's likely something else holding a filehandle open so we'd need to know more about what is holding onto the a filehandle on that filesystem. `lsof` or `fuser` should provide a bit more info.

williamh:
We are running github-backup-utils 2.11.1, which is the latest I see, and it does not solve the issue, so I will add the environment variable setting.

Let me know if we should keep this open.

lildude:
Lets leave this open for the moment as we still need to know what is holding a file handle open (this is just an assumption at the mo) just in case it _is_ backup-utils.

Does setting the env var solve the issue?

What does `fuser` or `lsof` show for the processes and files holding backup mount when you hit the problem? 

If this is SSH/backup-utils related, seeing this output will help us identify where the possible problem is.

williamh:
Yes, adding the following line to our backup script solved the issue:

```
export GHE_DISABLE_SSH_MUX=1
```

I don't have a log of running it without the environment variable on hand, but I  did check at the time 
and lsof showed an ssh process.

williamh:
I ran another backup today so I could get the lsof output. Here is what it shows:

```
[...]

PR #348, opened 2017-10-26
Switch to TMPDIR when working around OpenSSH bug

lildude:
As identified in https://github.com/github/backup-utils/issues/345, if we don't switch to `$TMPDIR` before working around the OpenSSH bug, we end up holding a file handle open on the backup directory until such time as the master connection timeout has been reached (10mins).  This poses a problem if the backup directory is an NFS mount that needs to be umounted when the backup has finished.

Fixes https://github.com/github/backup-utils/issues/345

🎩  tip to @zmedico for the fix.

/cc @github/backup-utils

### DIFF
--- a/share/github-backup-utils/ghe-ssh
+++ b/share/github-backup-utils/ghe-ssh
@@ -61,7 +61,10 @@ if [ -z "$GHE_DISABLE_SSH_MUX" ]; then
   controlpath="$TMPDIR/.ghe-sshmux-$(echo -n "$user@$host:$port" | git hash-object --stdin | cut -c 1-8)"
   opts="-o ControlMaster=auto -o ControlPath=\"$controlpath\" -o ControlPersist=10m -o ServerAliveInterval=10 $opts"
   # Workaround for https://bugzilla.mindrot.org/show_bug.cgi?id=1988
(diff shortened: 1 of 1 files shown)


share/github-backup-utils/ghe-ssh
```diff
@@ -61,7 +61,10 @@ if [ -z "$GHE_DISABLE_SSH_MUX" ]; then
   controlpath="$TMPDIR/.ghe-sshmux-$(echo -n "$user@$host:$port" | git hash-object --stdin | cut -c 1-8)"
   opts="-o ControlMaster=auto -o ControlPath=\"$controlpath\" -o ControlPersist=10m -o ServerAliveInterval=10 $opts"
   # Workaround for https://bugzilla.mindrot.org/show_bug.cgi?id=1988
-  [ -S $controlpath ] || ssh -f -p $port $opts -o BatchMode=yes "$host" -- /bin/true 1>/dev/null 2>&1 || true
+  if ! [ -S $controlpath ]; then
```
dbussink: Should we run the `cd` in a subshell so the working directory is not changed for the script itself? That side effect might trip up people in the future otherwise.

lildude: Ah, good point.

Smangele87 (commented):
test/test-ghe-restore.sh

PR #351, opened 2017-11-08
next iteration if there isn't an *.rsync file

donal:
If there happens to be zero alambic objects (unlikely on cluster, I know), you get this sort of error:

```
Backing up storage data ...
cat: /tmp/backup-utils-restore-GgXbCR/*.rsync: No such file or directory
* Transferring 0 objects from *
rsync: failed to open files-from file /tmp/backup-utils-restore-GgXbCR/*.rsync: No such file or directory
rsync error: syntax or usage error (code 1) at main.c(1453) [client=3.0.9]
```

And a failed backup:

```
Error: Snapshot incomplete. Some steps failed: alambic.
```

Fixed with this change—there could be better ways of doing this...should we just `break` instead of `continue`?


share/github-backup-utils/ghe-backup-alambic-cluster-ng
```diff
@@ -85,6 +85,7 @@ bm_end "$(basename $0) - Processing routes"
 # rsync all the repositories
 bm_start "$(basename $0) - Storage object sync"
 for file_list in $tempdir/*.rsync; do
+  [[ -f "$file_list" ]] || continue
```
lildude: Nit: we'd prefer to only [use `[[` for advanced bash functionality](https://github.com/github/backup-utils/blob/master/STYLEGUIDE.md#scripts-may-use--for-advanced-bash-features).

I also think it may be a good idea to only enter the loop if there is something to sync. I believe we should be able to guard against this by checking that `$routes_list` isn't empty. If it is, then there won't be any `$tempdir/*.rsync` files so we don't need to enter the loop.

donal: @lildude Thanks, that's the feedback I was looking for 😄  :+1:

lildude: I've got a local fix for this from some recent testing. Mind if I ~~trample all over~~ update your branch @donal?

donal: @lildude Thank you, please do! 🐘

donal:
> I like these changes now

Me too 😀

PR #354, opened 2017-11-28
test/test-ghe-backup.sh: skip test if not in a git checkout

austin987:
See https://bugs.gentoo.org/629628 for the background. In short, the test always assumed it was a git repo, and so git clone would always be available. If someone runs the tests from a tarball release, that's not true. The file is still there, just need to skip the clone.


austin987:
Nice catch, thanks. Fixed. Test now skips if .git is not present and usable.

PR #356, opened 2017-12-04
Docker image variable tests

djdefi:
Adds a few additional test items to the Docker build tests.

In adding a test to validate the correct ENV variables being set in `/etc/github-backup-utils/backup.config` , I realized I was using a bad `sed` and `tr` statement in the init script, which was leading to an invalid formatted / useless config within container.

This wasn't causing the backups to fail,as backup-utils was just reading the correctly formatted env `GHE_` variables set at runtime in the container.

I suppose this exposes the possibility of just removing `/etc/github-backup-utils/backup.config` and related function for writing the env variables there completely, but maybe it is good to have the fallback?


djdefi:
Pushed another commit to fix a test, could have had a false positive based on the Container exit status rather than the check I wanted.

PR #346, opened 2017-10-23
Docker container

djdefi:
Adds a Dockerfile for backup-utils stable.

( Below has been added as basis for documentation in `docs/` )

#### Setting backup configuration options for Docker
The `backup.config` file is dynamically populated at runtime with all `GHE_` environment variables that are part of the run command or Docker environment: 

```
$ docker run -it -e "GHE_HOSTNAME=hostname" \
-e "GHE_DATA_DIR=/data" \
-e "GHE_EXTRA_SSH_OPTS=-i /ghe-ssh/id_rsa -o UserKnownHostsFile=/ghe-ssh/known_hosts \
-e "GHE_NUM_SNAPSHOTS=15" \
-v "ghe-backup-data:/data" \
-v "$HOME/.ssh/known_hosts:/ghe-ssh/known_hosts" \
-v "$HOME/.ssh/id_rsa:/ghe-ssh/id_rsa" \
--rm \
backup-utils ghe-backup
```

It is also possible to specify a `-e GHE_BACKUP_CONFIG` flag and volume mount in a local `backup.config` file rather than specify the variables individually at run time:

```
$ docker run -it  -e "GHE_BACKUP_CONFIG=/mnt/backup.config" \
-v "ghe-backup-data:/data" \
-v "$HOME/.ssh/known_hosts:/ghe-ssh/known_hosts" \
-v "$HOME/.ssh/id_rsa:/ghe-ssh/id_rsa" \
-v "$HOME/backup-utils/backup.config:/mnt/backup.config" \
--rm \
backup-utils ghe-backup
```

#### SSH Keys with Docker

A SSH private key that has been added to the GitHub Enterprise [Management Console for administrative SSH access](https://help.github.com/enterprise/admin/guides/installation/administrative-shell-ssh-access/) needs to be mounted into the container from the host system. It is also recommended to mount a SSH `.ssh/known_hosts` file into the container.

```
$ docker run -it -e "GHE_HOSTNAME=hostname" \
-e "GHE_DATA_DIR=/data" \
-e "GHE_EXTRA_SSH_OPTS=-i /ghe-ssh/id_rsa -o UserKnownHostsFile=/ghe-ssh/known_hosts \
-v ghe-backup-data:/data \
-v "$HOME/.ssh/known_hosts:/ghe-ssh/known_hosts" \
-v "$HOME/.ssh/id_rsa:/ghe-ssh/id_rsa" \
--rm \
backup-utils ghe-backup
```

#### Managing backup data with Docker 

[...]

PR #360, opened 2017-12-13
More efficiently extract and quote environment variables when writing to config file on Docker

lildude:
I noticed yesterday that when using our instructions to pass the `GHE_EXTRA_SSH_OPTS` variable to a docker container, that the backup was failing:

```
$ docker run -it -e "GHE_HOSTNAME=10.0.1.64" \
-e "GHE_DATA_DIR=/data" \
-e "GHE_EXTRA_SSH_OPTS=-i /ghe-ssh/id_rsa -o UserKnownHostsFile=/ghe-ssh/known_hosts" \
-e "GHE_NUM_SNAPSHOTS=15" \
-v "ghe-backup-data:/data" \
-v "$HOME/.ssh/known_hosts:/ghe-ssh/known_hosts" \
-v "$HOME/.ssh/aws-lildude.pem:/ghe-ssh/id_rsa" \
-v "$(dirname $SSH_AUTH_SOCK):$(dirname $SSH_AUTH_SOCK)" \
-e "SSH_AUTH_SOCK=$SSH_AUTH_SOCK" \
--rm \
github/backup-utils ghe-backup
Starting backup of 10.0.1.64 in snapshot 20171213T154200
Warning: Identity file -o not accessible: No such file or directory.
ssh: Could not resolve hostname batchmode=no: Name or service not known
Error: ssh connection with '10.0.1.64' failed
Note that your SSH key needs to be setup on 10.0.1.64 as described in:
$
```

Note...

```
Warning: Identity file -o not accessible: No such file or directory.
ssh: Could not resolve hostname batchmode=no: Name or service not known
```

Whoopsie 😮 

After a lot of digging, I worked out that this was happening because of the way `ghe-docker-init` was taking the environment variables and writing them to file. The problem is the current method doesn't take into account variable values with spaces. These aren't being quoted which means each "word" is being treated as a different option and silently dropped after the first.

This PR addresses this by simplifying the process of writing the environment variables to file and quoting them all, like we suggest they are in the `backup-utils.config-example` file, to ensure we don't end up with inadvertent option splitting.

/cc @djdefi as the original implementor. I'm not sure how I missed this or how this worked when I tested https://github.com/github/backup-utils/pull/346 as it _did_ work for me AFAICR. 🤔

### DIFF
--- a/share/github-backup-utils/ghe-docker-init
+++ b/share/github-backup-utils/ghe-docker-init
@@ -6,12 +6,6 @@ mkdir -p /etc/github-backup-utils
 
 touch /etc/github-backup-utils/backup.config
 
-for VAR in $(env); do
-  if [[ $VAR =~ ^GHE_ ]]; then
-      backuputils_name=$(echo "$VAR" | sed -r "s/(.*)=.*/\1/g")
-      backuputils_value=$(echo "$VAR" | sed -r "s/.*=(.*)/\1/g")
-      echo "${backuputils_name}=${backuputils_value}" >> /etc/github-backup-utils/backup.config
-  fi
-done
+env | grep ^GHE_ | sed -r "s/(.[^=]+)=(.*)/\1=\"\2\"/g" >> /etc/github-backup-utils/backup.config
 
 exec "$@"
--- a/test/test-docker-build.sh
+++ b/test/test-docker-build.sh
@@ -41,7 +41,15 @@ begin_test "GHE_ env variables set in backup.config"
 (
   set -e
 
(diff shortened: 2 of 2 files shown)


djdefi:
Oops, good catch. I believe it worked before when you tested because the script was not setting the variables in the backup.config properly, so backup-utils was pulling the variables right out of the Docker environment.

When I switched the script to write the variables correctly, I missed the quoting there, causing this to break.

lildude:
Ah, that makes sense. Thanks for the clarification @djdefi. Now to twiddle our thumbs until the macOS build finished.

PR #363, opened 2018-01-08
Cleanup SSH multiplexing on exit

lildude:
Whilst experimenting with other changes, I encountered problems with an old SSH multiplexing session still running for the node in question and needed to manually clean it up so I could continue. The problems were my own fault but it did highlight that we aren't being good citizens and aren't cleaning up after ourselves.

This PR corrects this and automatically cleans up the SSH multiplexing when `ghe-backup` and `ghe-restore` exit rather than leaving it hanging around to naturally time out after 10mins.

I've also taken the opportunity to start implementing some of the simpler ShellCheck recommendations.

/cc @github/backup-utils

### DIFF
--- a/share/github-backup-utils/ghe-ssh
+++ b/share/github-backup-utils/ghe-ssh
@@ -7,7 +7,8 @@
 set -e
 
 # Bring in the backup configuration
-. $( dirname "${BASH_SOURCE[0]}" )/ghe-backup-config
+# shellcheck source=share/github-backup-utils/ghe-backup-config
+. "$( dirname "${BASH_SOURCE[0]}" )/ghe-backup-config"
 
 opts="$GHE_EXTRA_SSH_OPTS"
 while true; do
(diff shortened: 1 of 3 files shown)


share/github-backup-utils/ghe-ssh
```diff
@@ -70,4 +75,13 @@ fi
 $GHE_VERBOSE_SSH && set -x
 
 # Exec ssh command with modified host / port args and add nice to command.
-exec ssh -p $port $opts -o BatchMode=yes "$host" -- $GHE_NICE $GHE_IONICE "$@"
+if [ -z "$cleanup_mux" ]; then
+  # Exec ssh command with modified host / port args and add nice to command.
+  exec ssh -p $port $opts -o BatchMode=yes "$host" -- $GHE_NICE $GHE_IONICE "$@"
+else
+  if [ -z "$GHE_DISABLE_SSH_MUX" ]; then
```
juruen: It could be nice to move this second `if` to `elif` to have one less if branch.

PR #365, opened 2018-01-09
Fix flaky ghe-detect-leaked-ssh-keys tests

lildude:
Whilst working on https://github.com/github/backup-utils/pull/363 and https://github.com/github/backup-utils/pull/364 I noticed that the "ghe-detect-leaked-ssh-keys leaked keys in [old|current] snapshot" tests are very very flaky on our CI as of https://github.com/github/backup-utils/pull/361, but not Travis. 

That PR didn't introduce any changes that would directly affect the test, but it did add a new test which seems to have increased the likelihood of these tests overlapping with each other and with others that modify the `ghe-ssh-leaked-host-keys-list.txt` file used in testing. 

I think the problem is these changes aren't atomic under testing, so I've changed the script and test to use an environment variable, if set, or the contents of the original file.  The env var is only pre-set under testing.

I've also implemented a few more ShellCheck recommendations whilst I'm at it.

/cc @github/backup-utils

### DIFF
--- a/share/github-backup-utils/ghe-detect-leaked-ssh-keys
+++ b/share/github-backup-utils/ghe-detect-leaked-ssh-keys
@@ -45,9 +45,10 @@ if (ssh-keygen -E 2>&1 | head -1 |  grep -q 'option requires an argument'); then
 fi
 
 # Bring in the backup configuration
-. $( dirname "${BASH_SOURCE[0]}" )/ghe-backup-config
+# shellcheck source=share/github-backup-utils/ghe-backup-config
+. "$( dirname "${BASH_SOURCE[0]}" )/ghe-backup-config"
 
-fingerprint_blacklist=$(cat "$GHE_BACKUP_ROOT/share/github-backup-utils/ghe-ssh-leaked-host-keys-list.txt")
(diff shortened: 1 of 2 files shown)


share/github-backup-utils/ghe-detect-leaked-ssh-keys
```diff
@@ -45,9 +45,10 @@ if (ssh-keygen -E 2>&1 | head -1 |  grep -q 'option requires an argument'); then
 fi
 
 # Bring in the backup configuration
-. $( dirname "${BASH_SOURCE[0]}" )/ghe-backup-config
+# shellcheck source=share/github-backup-utils/ghe-backup-config
+. "$( dirname "${BASH_SOURCE[0]}" )/ghe-backup-config"
 
-fingerprint_blacklist=$(cat "$GHE_BACKUP_ROOT/share/github-backup-utils/ghe-ssh-leaked-host-keys-list.txt")
+FINGERPRINT_BLACKLIST="${FINGERPRINT_BLACKLIST:-$(cat "$GHE_BACKUP_ROOT/share/github-backup-utils/ghe-ssh-leaked-host-keys-list.txt")}"
```
sridharavinash: 👍

PR #366, opened 2018-01-15
Remove Precise Travis CI builds

lildude:
Updated: After taking on feedback, this PR only removes the obsolete Ubuntu Precise builds.

---
Originally:

macOS is not officially supported and waiting for Travis to run our tests can take many many hours queuing alone. Local dev should be sufficient testing.

Precise is dead and no longer receiving security updates except for those with deep pockets paying Canonical lots of money for "Extended security maintenance for advanced customers" so we probably don't need to keep testing on it either.

So what's the impact? The build one before this one took:

screen shot 2018-01-15 at 17 27 04

The one before this one is still running:

screen shot 2018-01-15 at 17 27 43

This one: 

screen shot 2018-01-15 at 17 29 01

The big times are almost exclusively waiting in the macOS queue.

/cc @github/backup-utils

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,24 +1,6 @@
 ---
 matrix:
   include:
-    - os: osx
-      before_install:
-        - brew update
-        - brew install gnu-tar
-        - brew install moreutils
-      script: make test
-    - os: linux
-      dist: precise
-      sudo: required
-      addons:
-        apt:
-          packages:
-            - devscripts
-            - debhelper
-            - moreutils
-            - fakeroot
-      before_script: git fetch --unshallow
-      script: debuild -uc -us
     - os: linux
       dist: trusty
       sudo: required
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,12 @@
 ---
 matrix:
   include:
+    - os: osx
+      before_install:
+        - brew update
+        - brew install gnu-tar
+        - brew install moreutils
+      script: make test
     - os: linux
       dist: trusty
       sudo: required


dmittman:
macOS support is useful, and generally only requires minor modifications. As long as you're willing to entertain pull requests that maintain macOS compatibility, I'd be okay doing without a macOS CI build. Any idea why the macOS builds are taking so long?

lildude:
> As long as you're willing to entertain pull requests that maintain macOS compatibility

Definitely. It's what most of us use to dev backup-utils 😄

> Any idea why the macOS builds are taking so long?

Yeah, Travis doesn't have the capacity for the demand for public repo builds and can't really scale it
 like they can Linux builds so we'll always have a backlog to contend with 😞 . Ref:  https://github.com/travis-ci/travis-ci/issues/7304#issuecomment-287591291 

Looking at https://www.traviscistatus.com/ it looks like they're limited to 90 concurrent macOS builds with a backlog that grows beautifully during :us: waking hours 😄

screen shot 2018-01-15 at 18 05 21

I'm happy to add macOS back if others feel it needs to remain.

rubiojr:
> I'm happy to add macOS back if others feel it needs to remain.

@lildude 👍 to leave it around. If we need to quickly merge PRs (release PRs, critical fixes, etc), we could always ignore them.

lildude:
@rubiojr macOS builds are back. Only Precise builds are removed now.

lildude:
I'm going to go ahead and merge this without waiting for the final macOS build to run. The travis-ci/pr job includes the same tests and finally ran after queuing for ~6h20m. The travis-c/push job has been queuing for over 7h30m already 😱

PR #369, opened 2018-02-12
Optimise route generation and finalisation during cluster restores of pages

lildude:
When the cluster-related restores were optimised in https://github.com/github/backup-utils/pull/322 we missed the Pages restore. This PR implements the same optimisation for Page restores.

Depends on https://github.com/github/github/pull/85086

/cc @github/backup-utils


lildude:
https://github.com/github/github/pull/85086 has been merged now. Review please @github/backup-utils or @github/for-business-vm

PR #374, opened 2018-03-08
Use primary index size to determine if an index has changed

snh:
This PR updates the method we use to track which Audit Log and Hookshot Log Elasticsearch indices have changed since the previous backup, and therefore need to be retransferred.

Previously we considered historic indices (previous months in the case of audit logs, and days in the case of hookshot logs) as immutable, and therefore didn't retransfer these on subsequent backups once a final backup of the index had been completed at the end of the relevant period.

While this immutability remains the case in most circumstances, there are occasionally situations where historic indices may have changed, such as when the format of the index changes, or information needs to be scrubbed.

To identify when a historic index has changed in any way, and therefore needs to be retransferred, we now keep track of the size of the indices in bytes.


share/github-backup-utils/ghe-backup-es-audit-log
```diff
@@ -26,31 +26,34 @@ else
...
+
 # Determine if the audit log migration has occurred or is needed.
 if echo 'set -o pipefail; ! test -e /data/user/common/es-scan-complete && test -f /usr/local/share/enterprise/run-audit-log-transitions.sh' | ghe-ssh "$host" /bin/bash; then
   if echo 'set -o pipefail; echo n | /usr/local/share/enterprise/run-audit-log-transitions.sh > /dev/null 2>&1 && touch /data/user/common/es-scan-complete' | ghe-ssh "$host" /bin/bash; then
     touch $GHE_SNAPSHOT_DIR/es-scan-complete
   fi
 fi
 
-current_index="audit_log(-[0-9]+)?-$(ghe-ssh "$host" 'date +"%Y-%m"')(-[0-9]+)?"
+while read index; do
```
lildude: Any particular reason for the switch from a `for` loop to a `while`?  Curious as I nearly missed this was reading in `$indices` as it's right at the end.

share/github-backup-utils/ghe-backup-es-audit-log
```diff
@@ -26,31 +26,34 @@ else
...
+  set $index
+  index_name=$1
+  index_size=$2
 
-for index in $indices; do
-  if [[ -f $GHE_DATA_DIR/current/audit-log/$index.gz && -f $GHE_DATA_DIR/current/audit-log/$index.gz.complete && ! $index =~ $current_index ]]; then
-    # Hard link any older indices that are complete, since these won't change
-    ln $GHE_DATA_DIR/current/audit-log/$index.gz $GHE_SNAPSHOT_DIR/audit-log/$index.gz
-    ln $GHE_DATA_DIR/current/audit-log/$index.gz.complete $GHE_SNAPSHOT_DIR/audit-log/$index.gz.complete
```
lildude: As we're switching methods, I think it's worth cleaning up all the old and now unused `*.complete` files here whilst we're at it. They take no space, but will unnecessarily occupy inodes once this PR ships.

share/github-backup-utils/ghe-backup-es-hookshot
```diff
@@ -20,21 +20,33 @@ ghe_remote_version_required "$host"
...
-for index in $indices; do
-  if [ -f $GHE_DATA_DIR/current/hookshot/$index.gz -a -f $GHE_DATA_DIR/current/hookshot/$index.gz.complete -a $index \< $current_index ]; then
-    # Hard link any older indices that are complete, since these won't change
-    ln $GHE_DATA_DIR/current/hookshot/$index.gz $GHE_SNAPSHOT_DIR/hookshot/$index.gz
-    ln $GHE_DATA_DIR/current/hookshot/$index.gz.complete $GHE_SNAPSHOT_DIR/hookshot/$index.gz.complete
+if [ $GHE_VERSION_MAJOR -ge 2 ] && [ $GHE_VERSION_MINOR -ge 2 ]; then
+  es_port=9201
+else
+  es_port=9200
+fi
```
lildude: We weren't making this differentiation before, and won't need it soon. Can we remove this and put 👇 back to the way it was before.

share/github-backup-utils/ghe-backup-es-hookshot
```diff
@@ -20,21 +20,33 @@ ghe_remote_version_required "$host"
...
+
+if ! indices=$(ghe-ssh "$host" "curl -s \"localhost:$es_port/_cat/indices/hookshot-logs-*?h=index,pri.store.size&bytes=b\""); then
+  echo "Error: failed to retrieve hookshot log indices." 1>&2
+  exit 1
+fi
+
+# Hookshot indices may not exist if no recent webhook deliveries have occured.
+[ -z "$indices" ] && exit
+
+while read index; do
```
lildude: Ditto comment about `for` to `while` switch as above.

share/github-backup-utils/ghe-backup-es-hookshot
```diff
@@ -20,21 +20,33 @@ ghe_remote_version_required "$host"
...
+
+  if [[ -f $GHE_DATA_DIR/current/hookshot/$index_name.gz && $(cat $GHE_DATA_DIR/current/hookshot/$index_name.gz.size 2>/dev/null || true) -eq $index_size ]]; then
+    # Hard link any indices that have not changed since the last backup
+    ln $GHE_DATA_DIR/current/hookshot/$index_name.gz $GHE_SNAPSHOT_DIR/hookshot/$index_name.gz
+    ln $GHE_DATA_DIR/current/hookshot/$index_name.gz.size $GHE_SNAPSHOT_DIR/hookshot/$index_name.gz.size
   else
-    ghe-ssh "$host" "/usr/local/share/enterprise/ghe-es-dump-json \"http://localhost:9201/$index\"" | gzip > $GHE_SNAPSHOT_DIR/hookshot/$index.gz
-    if [ $index \< $current_index ]; then
-      touch $GHE_SNAPSHOT_DIR/hookshot/$index.gz.complete
-    fi
```
lildude: Ditto comment about cleaning up the `*.complete` files as above.

share/github-backup-utils/ghe-backup-es-hookshot
snh: I added this in for a couple of reasons:

- To keep it consistent with the audit log script.
- As I now catch errors when retrieving the hookshot log indices, we need to fall back to 9200 in case we are running on GHE 2.0 or 2.1, which doesn't have Elasticsearch available on 9201. An alternative would be to not run `ghe-backup-es-hookshot` on these older versions since hookshot logs aren't in Elasticsearch there, but this would involve adding additional version logic to `bin/ghe-backup`.

share/github-backup-utils/ghe-backup-es-hookshot
snh: I didn't worry about this for a couple of reasons:

- These will cycle out eventually, as only older snapshots will contain these.
- Modifying previous backup snapshots is something we have generally avoided (apart from pruning them completely), so I didn't want to set a new precedent.

share/github-backup-utils/ghe-backup-es-audit-log
snh: I went with a `while` to avoid the need to switch `IFS` back and forth. I've updated to use a `for` loop now as is it more readable as you mentioned.

share/github-backup-utils/ghe-backup-es-audit-log
snh: I didn't worry about this for a couple of reasons:

These will cycle out eventually, as only older snapshots will contain these.
Modifying previous backup snapshots is something we have generally avoided (apart from pruning them completely), so I didn't want to set a new precedent.

share/github-backup-utils/ghe-backup-es-hookshot
lildude: 👍 make sense in the current context.

Issue #371, opened 2018-02-27
Release for GitHub Enterprise 2.12+?


mdwheele:
Hello!

The [Backups and disaster recovery](https://help.github.com/enterprise/2.12/admin/guides/installation/backups-and-disaster-recovery/) documentation for GitHub Enterprise 2.12 instruct to use the "latest release" of `backup-utils`. It seems that historically, the release versions have followed closely with the supported version of GitHub Enterprise. 

Is this simply a coincidence that they happen to have matched? We're running `backup-utils@v2.11.2` on our instance (2.12.5) and things seem to be rolling fine. I have faint memory of reading that Enterprise will always support one version behind current for `backup-utils`, but I can't find that now. 

Anywho, I'm more or less looking for some clarification on how the release process works. If it can be explicated here, I'd be willing to contribute some documentation if you think it'd be helpful.

lildude:
Yes, for the most part is has been a coincidence that the version numbers have matched historically. We've tended to only bump the major version in line with the GHE release if there is a requirement for it in that GHE release, hence we've skipped entire major versions in the past.

That said, we are due another patch release which I plan to do tomorrow. The next major release will likely coincide with the 2.13 release and will be quite significant, with a significant improvement in the documentation too.

Regarding support, the two versions thing is regarding the number of "steps" we support for upgrades and restores of GHE as opposed to backup-utils itself eg we'll support upgrading from 2.9 directly to 2.11 but not to 2.12 as that is three steps. The same thing for restores: you can restore a backup of a 2.9 instance to 2.11 but not 2.12, regardless of the version of backup-utils provided that version supports both versions.

mdwheele:
Sounds good. So in theory, staying current with latest release of Enterprise and `backup-utils` should be good enough then?

lildude:
> Sounds good. So in theory, staying current with latest release of Enterprise and backup-utils should be good enough then?

Yup. If a specific release of backup-utils is ever needed, it'll be in the Enterprise release notes for that release and also forced (and reported if not met) through the version check that happens early in the backup and restore process.

lildude:
To close this out, here's the note I added to the [2.13.0 release](https://github.com/github/backup-utils/releases/tag/v2.13.0) as an indicator of the way we'll be doing things in future:

[...]

PR #375, opened 2018-03-10
WIP: Unify the backup & restore process

lildude:
## Background

Since the introduction of Cluster, backup-utils has effectively maintained three different methods of backing up GitHub Enterprise appliances, whilst at the same time attempted to retain backwards compatibility for all releases of GitHub Enterprise since 11.10.344.

This has resulted in a very messy, hard to maintain and not particularly efficient code base, with the cluster-specific backup functionality not being tested within backup-utils at all.  This PR starts to address this by unifying the backup methods used to a single method that is compatible with cluster and single nodes. Switching to this method will effectively be testing the same pathways used for cluster with the need to only write a few more cluster-specific tests for those scenarios not covered by testing for single nodes.

## Caveats

The biggest caveat with this change is backup-utils 2.13.0 onwards will only support GitHub Enterprise 2.11.0 and newer as these are the only releases of GitHub Enterprise that contain the necessary functionality.

Prior to this PR being merged, a new branch will be created to allow for easy (manual) maintenance of legacy releases of backup-utils that people may be using to backup 2.10 and older releases. Previous releases will _not_ be removed.

## The Future

The intention going forward is to keep the versioning and support for backup-utils releases inline with that of GitHub Enterprise to make it clearer what versions of backup-utils are supported and with which versions of GitHub Enterprise, hopefully addressing concerns like those raised in https://github.com/github/backup-utils/issues/371 .

PRs are in the pipeline for further beautification of the code and completely updating and reformatting the README and documentation. These changes haven't been included in this PR as they detract from the primary purpose of this PR and would turn what is already a pretty large PR into a huge and hard to review PR.

## What does this PR do?

In short, this PR:

- Removes all logic and support for legacy 11.10.xxx GitHub Enterprise releases and GitHub Enterprise releases prior to 2.11.0.
- Switches to a single method for backing single nodes and clusters, based on the code used to backup clusters.
- Unifies the format of the remaining code putting things in a better position for further simplification in future.
- Removes unused commands and tests.
- Adds more tests to cover more of the remaining code.
- Removes duplication within the tests and code.

## TODO

- [ ] More 👀 on these changes.
- [ ] More tests.
- [ ] Extensive manual testing with various configuration combinations to make sure I've not missed anything and write tests for those things I have.
- [ ] Make this work with restoring to unconfigured instances - pending changes on the GitHub Enterprise side of things.


share/github-backup-utils/ghe-backup-storage
```diff
@@ -0,0 +1,129 @@
...
+
+# Make sure root backup dir exists if this is the first run
+mkdir -p "$backup_dir"
+
+# Removes the remote sync-in-progress file on exit, re-enabling GC operations
+# on the remote instance.
+cleanup() {
+  # Enable remote maintenance operations
+  for hostname in $hostnames; do
+    ghe-gc-enable $ssh_config_file_opt $hostname:$port
```
sbryant: if ssh fails on any of these the rest of the operations will fail? Maybe echo the failure to enable gc operations on the specified host?

lildude: Good idea. I'll do this elsewhere too.

bin/ghe-restore
```diff
@@ -108,39 +108,27 @@ ghe_remote_version_required "$GHE_HOSTNAME"
...
 if ghe-ssh "$GHE_HOSTNAME" -- \
   "[ -f '$GHE_REMOTE_ROOT_DIR/etc/github/cluster' ]"; then
-  cluster=true
+  CLUSTER=true
 fi
+export CLUSTER
 
 # Restoring a cluster backup to a standalone appliance is not supported
-if ! $cluster && [ "$GHE_BACKUP_STRATEGY" = "cluster" ]; then
+if ! $CLUSTER && [ "$GHE_BACKUP_STRATEGY" = "cluster" ]; then
```
snh: This currently blocks the restoration of a backup taken from a HA pair to a standalone appliance.

lildude: Should be fixed now, and with a test too.

lildude:
@github/backup-utils this is now ready for review. The one failing test is due to `parallel` not being installed on the new CI servers which I'm working to address.

share/github-backup-utils/ghe-restore-repositories
```diff
@@ -0,0 +1,177 @@
...
+
+# Tell dgit about the repositories restored
+if $CLUSTER; then
+  bm_start "$(basename $0) - Finalizing routes"
+  ghe_verbose "Finalizing routes"
+  cat $to_restore | ghe-ssh "$GHE_HOSTNAME" -- sponge $to_restore
+  ghe-ssh "$GHE_HOSTNAME" -- /bin/bash >&3 <<EOF
+    split -l 1000 $to_restore $tempdir/chunk
+    chunks=\$(find $tempdir/ -name chunk\*)
+    echo "\$chunks" | xargs -P64 -n1 -I{} /bin/sh -c "cat {} | github-env ./bin/dgit-cluster-restore-finalize"
```
lildude: Heads up @snh. Since you last looked at this, I've switched out the use of `parallel` in favour of `xargs -P64` for all of the finalize steps (with a few force pushes to hide my mess 😊). 

My testing shows this works just as well, but I'd appreciate it if you could specifically double check me here - I'm not sure why you went for `parallel` over `xargs` when you first implemented this. 🙇

snh: @lildude I'm concerned that 64 will be a little high in some circumstances, as there are a number of Cluster customers who restore to DR and test clusters that are quite light-on resources wise.

From my quick testing, each `dgit-cluster-restore-finalize`, without having been feed any routes yet, uses 76MB of PSS memory, which adds up to close to 5GB for 64 of these.

For this reason I'm quite a fan of the approach we were using with `parallel` to scale the number of processes based on the CPU count, which is often a good indicator of the resources a node has available to it.

Did you have a chance to try and stub `parallel`? As this runs on the appliance itself, which already had `parallel`, the dependency on `parallel` has never really been an issue until now.

lildude: > I'm concerned that 64 will be a little high in some circumstances, as there are a number of Cluster customers who restore to DR and test clusters that are quite light-on resources wise.

This would only affect customers who actually have up to 64 CPUs on their node as `-P` is for maximum number of processes. From the man page (emphasis is mine): "Run  **up  to**  max-procs processes at a time;". 

I'd hope if they're going to assign that many CPUs, they'd allocate an equivalently large amount of memory too as they'd hit other problems long before they hit backup-utils related problems.

> For this reason I'm quite a fan of the approach we were using with parallel to scale the number of processes based on the CPU count

I'd have liked to do the same with `xargs` by not specifying a number or using 0 to use all available procs, but `xargs` on macOS requires a value > 0 and we still test on macOS, hence I've gone for a rather large value to mimic the same behaviour.

> Did you have a chance to try and stub parallel? 

I did, but this turned out to be a lot more complex than anticipated and I soon found myself effectively writing a passthrough version of `parallel` that stripped the args so we could run the subsequent commands. This adds unnecessary complexity and potential tech debt, just for testing, when a simple and just as effective alternate solution works for both actual use and testing.

> As this runs on the appliance itself, which already had parallel, the dependency on parallel has never really been an issue until now.

That's because we were never testing this behaviour within backup-utils itself so never encountered any problems until now. We don't use `parallel` for anything else within GitHub Enterprise either.

lildude: As discussed in Slack, I'm wrong here and misunderstood the behaviour. Will revert that change.

lildude:
Just to be 100% clear about the biggest change in this PR, I'm going to state it in a comment of its own, using a mix of @snh's and my own words:

The backup utilities will now only support the current feature release and two previous feature releases. For example, the backup utilities 2.13.0 will support 2.11, 2.12 and 2.13. A new legacy branch will be created, and maintained with critical bug fixes, for customers using older releases.

A new backup-utils release will be made with each GitHub Enterprise feature release in future so backup-utils 2.14 will support GitHub Enterprise 2.12, 2.13 and 2.14.

The documentation is being updated to reflect this change in https://github.com/github/backup-utils/pull/380

PR #376, opened 2018-03-13
Add SSH version 5.6 requirement

mwiesen:
OpenSSH 5.6 was the first version to [support the `ControlPersist` option](https://www.openssh.com/txt/release-5.6), which ghe-ssh relies on when enabling multiplexing. This change calls out the requirement to avoid any confusion when running backup-utils on older systems.


jatoben:
LGTM! Thanks for catching this, I overlooked it in the earlier PR (#368). cc @github/backup-utils for review.

PR #377, opened 2018-03-13
Retry with the admin ssh port on network unreachable too.

blovett:
Retry with the admin ssh port when a network unreachable error is received. This is necessary in environments where direct ssh is restricted under normal circumstances.


lildude:
Gonna merge this now. Thanks for the PR @blovett.

PR #381, opened 2018-03-26
Update post-restore messaging

keeran:
Following a restore we should prompt the user to save settings on the appliance as this will trigger any necessary migrations.

### DIFF
--- a/bin/ghe-restore
+++ b/bin/ghe-restore
@@ -448,8 +448,8 @@ else
   ghe-ssh "$GHE_HOSTNAME" -- /bin/sh 1>&3
 fi
 
-echo "Completed restore of $GHE_HOSTNAME from snapshot $GHE_RESTORE_SNAPSHOT"
+echo "Finished restore of $GHE_HOSTNAME from snapshot $GHE_RESTORE_SNAPSHOT"
 
 if ! $cluster; then
-  echo "Visit https://$hostname/setup/settings to review appliance configuration. Settings must be applied to complete any necessary migrations."
+  echo "To complete the restore process, please visit https://$hostname/setup/settings to review and save the appliance configuration."
 fi
--- a/bin/ghe-restore
+++ b/bin/ghe-restore
@@ -451,5 +451,5 @@ fi
 echo "Completed restore of $GHE_HOSTNAME from snapshot $GHE_RESTORE_SNAPSHOT"
 
 if ! $cluster; then
-  echo "Visit https://$hostname/setup/settings to review appliance configuration."
+  echo "Visit https://$hostname/setup/settings to review appliance configuration. Settings must be applied to complete any necessary migrations."
 fi
--- a/bin/ghe-restore
+++ b/bin/ghe-restore
@@ -450,6 +450,6 @@ fi
 
 echo "Restore of $GHE_HOSTNAME from snapshot $GHE_RESTORE_SNAPSHOT finished."
 
-if ! $cluster; then
+if ! $instance_configured; then
   echo "To complete the restore process, please visit https://$hostname/setup/settings to review and save the appliance configuration."
 fi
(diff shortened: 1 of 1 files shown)


bin/ghe-restore
```diff
@@ -451,5 +451,5 @@ fi
 echo "Completed restore of $GHE_HOSTNAME from snapshot $GHE_RESTORE_SNAPSHOT"
 
 if ! $cluster; then
-  echo "Visit https://$hostname/setup/settings to review appliance configuration."
+  echo "Visit https://$hostname/setup/settings to review appliance configuration. Settings must be applied to complete any necessary migrations."
```
tjl2: How about:

> You must now visit https://$hostname/setup/settings to review and save appliance configuration

I feel like this is giving a direct command to the user and actually mentions 'saving' the configuration.  Users who are unfamiliar with GHE may not know that hitting 'Save settings' is the equivalent of 'applying settings'?

lildude: I agree we should probably be a little firmer here, though "you must" feels a little odd to me.

How about:

> Please visit https://$hostname/setup/settings to review and save the appliance configuration.

tjl2: I still feel that sounds almost optional though… How about:

> To complete the restore process, please visit https://$hostname/setup/settings to review and save the appliance configuration.

Then at least we're giving people the reason for doing this.

lildude: Oooo, nice ✨

keeran: Hahah words are hard. Hesitated making this change due to the repetition of 'complete' (previous output says it's complete, now I need to do something to complete it). If neither of you think it's an issue I'll go ahead with it :)

lildude: Good point @keeran. How about changing "Completed" in the line above to "Finished", so users will see:

```
Finished restore of 5.5.5.5:122 from snapshot 20180326T020444
To complete the restore process, please visit https://5.5.5.5/setup/settings to review and save the appliance configuration.
```

Or move to the end of the line:

```
Restore of 5.5.5.5:122 from snapshot 20180326T020444 finished.
To complete the restore process, please visit https://5.5.5.5/setup/settings to review and save the appliance configuration.
```

tjl2: @keeran @lildude good suggestions :thumbsup:

My vote goes for the second one:

```
Restore of 5.5.5.5:122 from snapshot 20180326T020444 finished.
To complete the restore process, please visit https://5.5.5.5/setup/settings to review and save the appliance configuration.
```

keeran: Thanks chaps! ✨

bin/ghe-restore
```diff
@@ -448,8 +448,8 @@ else
   ghe-ssh "$GHE_HOSTNAME" -- /bin/sh 1>&3
 fi
 
-echo "Completed restore of $GHE_HOSTNAME from snapshot $GHE_RESTORE_SNAPSHOT"
+echo "Finished restore of $GHE_HOSTNAME from snapshot $GHE_RESTORE_SNAPSHOT"
 
 if ! $cluster; then
-  echo "Visit https://$hostname/setup/settings to review appliance configuration."
+  echo "To complete the restore process, please visit https://$hostname/setup/settings to review and save the appliance configuration."
```
snh: It should only be necessary to do this if you are restoring on to an unconfirmed appliance, as we automatically run `ghe-config-apply` on configured appliances at the end of the restore process: https://github.com/github/backup-utils/blob/7d0ca70cd3df6d1d0f3dd5d6ff6ded81a6db5c08/bin/ghe-restore#L400-L413

Maybe the `if` could be changed to `if ! $instance_configured; then` so that this message only appears on un-configured appliances.

lildude: Ah, good point. Nice catch @snh.

lildude: @keeran don't forget this 😉 (I'd like to merge this before merging by big PR so it makes it in before I create the legacy branch).

keeran: @lildude ooops - read pre-coffee, thanks :)

PR #386, opened 2018-04-01
Treat missing repository networks, gists, and storage objects as a non-critical error 

snh:
With the recent move to the cluster backup methodology for the single instance and high availability architectures, the backup process has become less tolerant of data consistency issues.

In particular, if a repository network, gist, or storage object has a route in the database, but cannot be found on disk, the backup will abort, leaving an unusable backup snapshot.

While such errors are serious and warrant investigation with assistance from GitHub Enterprise Support, I propose that we should continue on with the remainder of the backup regardless, and present a warning regarding these routes. This can be accomplished using the `--ignore-missing-args` command-line argument for `rsync`.

To go on to detect these inconsistencies in a more parsable manner, a _Verifying Routes_ process has been added to `ghe-backup-repositories` and `ghe-backup-storage`. This compares the routes that were received from GitHub Enterprise with the data that ended up in the backup snapshot, and reports any differences that are found as a warning, along with a recommendation to contact GitHub Enterprise Support.

As `--ignore-missing-args` was only introduced in `rsync` 3.1.0, if an earlier version is in use I instead catch exits with an exit code of `23`, which is returned for a number of errors including the `rsync: link_stat <file> failed: No such file or directory (2)` error.

~In the future I'd like to swap the use of `diff` out for `git diff -- <file1> <file2>`. This is dependent on `git` becoming mandatory for the backup utilities.~ Edit: Done.

/cc @github/backup-utils @jatoben


lildude:
> As --ignore-missing-args was only introduced in rsync 3.1.0, if an earlier version is in use I instead catch exits with an exit code of 23, which is returned for a number of errors including the rsync: link_stat <file> failed: No such file or directory (2) error.

Looks like this needs work as it appears to be what is failing on the macOS build:

```
 * Transferring auxiliary files ...
    rsync: --ignore-missing-args: unknown option
    rsync error: syntax or usage error (code 1) at /BuildRoot/Library/Caches/com.apple.xbs/Sources/rsync/rsync-51/rsync/main.c(1337) [client=2.6.9]
```

> In the future I'd like to swap the use of diff out for git diff -- <file1> <file2>. This is dependent on git becoming mandatory for the backup utilities.

This is already a [documented requirement](https://github.com/github/backup-utils/blob/master/docs/requirements.md#backup-host-requirements) so you could go this route and have the normal `diff` as a fallback.

More comments coming in my review.

share/github-backup-utils/ghe-backup-repositories
```diff
@@ -351,4 +353,13 @@ RULES
 done
 bm_end "$(basename $0) - Special Data Directories Sync"
 
+bm_start "$(basename $0) - Verifying Routes"
+
+cat $tempdir/*.rsync | sort | uniq > $tempdir/source_routes
+(cd $backup_dir/ && find * -mindepth 6 -maxdepth 7 -name \*.git -exec dirname {} \; | sort | uniq) > $tempdir/destination_routes
+
+diff $tempdir/source_routes $tempdir/destination_routes || echo "Warning: One or more repository networks and/or gists were not found on the source appliance. Please contact GitHub Enterprise Support for assistance."
```
lildude: Several points:

- I think we should only show the difference output when a verbose backup is performed. 

- Do you think we should also force the backup to exit non-zero if this difference is found so it doesn't go completely under the radar? ie allow the rsync to continue but still "fail" the backup step.

- Is this going to have a major impact on the backup time, particularly for larger datasets?

share/github-backup-utils/ghe-rsync
```diff
@@ -14,14 +14,26 @@ set -o pipefail
 # stderr (rsync versions >= 3.x). The complex redirections are necessary to
 # filter stderr while also keeping stdout and stderr separated.
 IGNOREOUT='^(file has vanished: |rsync warning: some files vanished before they could be transferred)'
+
+# Remove --ignore-missing-args argument if using an earlier version of rsync.
+if [[ $GHE_EXTRA_RSYNC_OPTS == *"--ignore-missing-args"* ]] && ! rsync -h | grep -q '\-\-ignore-missing-args'; then
+  GHE_EXTRA_RSYNC_OPTS=${GHE_EXTRA_RSYNC_OPTS//--ignore-missing-args/}
```
lildude: This is going to need to be extended to strip `--ignore-missing-args` from the list of arguments passed in too else we hit the failure seen in the current macOS builds.

snh: Good catch, I was looking for `--ignore-missing-args` in the wrong place!

share/github-backup-utils/ghe-backup-repositories
snh: > I think we should only show the difference output when a verbose backup is performed.

As this information will be needed by Enterprise Support, I'd rather leave it in the standard output to prevent the need to re-run the backup with the `-v` flag. As this output forms part of the warning, I wouldn't consider it informational.

> Do you think we should also force the backup to exit non-zero if this difference is found so it doesn't go completely under the radar? ie allow the rsync to continue but still "fail" the backup step.

This seems like a sensible approach, and would certainly add additional visibility. I'll get this added in.

> Is this going to have a major impact on the backup time, particularly for larger datasets?

I don't expect it will have a significant impact, as I have tried to keep the comparison as lean and efficient as possible, but I do have a large backup archive here that I plan to use for some benchmarking.

snh: > This seems like a sensible approach, and would certainly add additional visibility. I'll get this added in.

On second thoughts, this probably isn't ideal, as it leaves the snapshot marked as incomplete, which will cause it to get pruned on the next successful backup.

snh:
> This is already a documented requirement so you could go this route and have the normal diff as a fallback.

PRs such as https://github.com/github/backup-utils/pull/362 and https://github.com/github/backup-utils/pull/378 gave me the impression we were trying to avoid a hard requirement for `git`?

> How about not implementing the --ignore-missing-args flag at all (removes the version dependency), and instead catch error 23 and use it to warn of a problem.

My concern with this approach is that exit code 23 appears to be used for a number of _partial transfer due to error_ related error states, some of which we don't want to ignore/silence.

share/github-backup-utils/ghe-backup-repositories
lildude: > On second thoughts, this probably isn't ideal, as it leaves the snapshot marked as incomplete, which will cause it to get pruned on the next successful backup.

Ah, good point.

lildude:
> PRs such as #362 and #378 gave me the impression we were trying to avoid a hard requirement for git?

Yup, that's why I suggested keeping the normal diff as a fallback, but I don't mind either way.

> My concern with this approach is that exit code 23 appears to be used for a number of partial transfer due to error related error states, some of which we don't want to ignore/silence.

Hmmm, but aren't we going to be doing this for those backup hosts _not_ running a new enough version of rsync?

test/test-ghe-backup.sh
```diff
@@ -298,3 +298,25 @@ begin_test "ghe-backup cluster"
...
+
+    if ! ghe-backup -v > "$TRASHDIR/backup-out" 2>&1; then
+      cat "$TRASHDIR/backup-out"
+      : ghe-backup should have completed successfully
+      false
+    fi
+
+    # Check the output for the warnings
+    cat "$TRASHDIR/backup-out" | grep "Warning: One or more repository networks and/or gists were not found on the source appliance."
+    cat "$TRASHDIR/backup-out" | grep "Warning: One or more storage objects were not found on the source appliance."
```
lildude: Think we should test to confirm the diff output is here too?

snh:
> Hmmm, but aren't we going to be doing this for those backup hosts not running a new enough version of rsync?

Indeed, though I am hopeful that most users will be running a more recent version of `rsync`. The route verification should help safeguard us against incomplete backups too.

share/github-backup-utils/ghe-rsync
```diff
@@ -13,15 +13,30 @@ set -o pipefail
...
-IGNOREOUT='^(file has vanished: |rsync warning: some files vanished before they could be transferred)'
-(rsync "${@}" $GHE_EXTRA_RSYNC_OPTS 3>&1 1>&2 2>&3 3>&- |
-  (egrep -v "$IGNOREOUT" || true)) 3>&1 1>&2 2>&3 3>&- |
-  (egrep -v "$IGNOREOUT" || true)
+ignoreout='^(file has vanished: |rsync warning: some files vanished before they could be transferred)'
+
+# Remove --ignore-missing-args parameter if using an earlier version of rsync.
+if ! rsync -h | grep -q '\-\-ignore-missing-args'; then
+  for parameter; do
+    [[ ! $parameter == "--ignore-missing-args" ]] && parameters+=("$parameter") || ignore23=1
```
lildude: Nit: https://github.com/github/backup-utils/blob/master/STYLEGUIDE.md#scripts-should-not-use-bash-arrays

snh: Still trying to find an alternative to this.

So far on all my attempts, if I try and convert the arguments to a string so to then remove `--ignore-missing-args`, it is stripping the quotes out, so things like `--rsync-path='sudo -u git rsync'` become `--rsync-path=sudo -u git rsync`.

lildude: For the life of me, I can't find where I've dealt with this situation in the past and can't think of a solution without resorting to array manipulation or getting into hacking brittle and confusing POSIX compliant solutions.

I think we can make an exception in this situation as we only want to remove an element from the `$@` array.

That said, I'm not sure this is doing as it should be when run on hosts _with_ support for this option:

```
$ rsync -h | grep '\-\-ignore-missing-args'
     --ignore-missing-args   ignore missing source args without error
$
$ bash -x ./share/github-backup-utils/ghe-rsync --rsync-path=\'sudo -u git rsync\' --ignore-missing-args
[...]
+ ignoreout='^(file has vanished: |rsync warning: some files vanished before they could be transferred)'
+ rsync -h
+ grep -q '\-\-ignore-missing-args'
+ for parameter in '"$@"'   <---- **I DONT THINK WE SHOULD BE HERE**
+ [[ ! --rsync-path='sudo == \-\-\i\g\n\o\r\e\-\m\i\s\s\i\n\g\-\a\r\g\s ]]
+ parameters+=("$parameter")
+ for parameter in '"$@"'
+ [[ ! -u == \-\-\i\g\n\o\r\e\-\m\i\s\s\i\n\g\-\a\r\g\s ]]
+ parameters+=("$parameter")
+ for parameter in '"$@"'
+ [[ ! git == \-\-\i\g\n\o\r\e\-\m\i\s\s\i\n\g\-\a\r\g\s ]]
+ parameters+=("$parameter")
+ for parameter in '"$@"'
+ [[ ! rsync' == \-\-\i\g\n\o\r\e\-\m\i\s\s\i\n\g\-\a\r\g\s ]]
+ parameters+=("$parameter")
+ for parameter in '"$@"'
+ [[ ! --ignore-missing-args == \-\-\i\g\n\o\r\e\-\m\i\s\s\i\n\g\-\a\r\g\s ]]
+ ignore23=1
+ egrep -v '^(file has vanished: |rsync warning: some files vanished before they could be transferred)'
+ rsync '--rsync-path='\''sudo' -u git 'rsync'\'''  <---- **THE ARG HAS GONE AWOL**
+ egrep -v '^(file has vanished: |rsync warning: some files vanished before they could be transferred)'
rsync: link_stat "/Users/lildude/github/backup-utils/git" failed: No such file or directory (2)
rsync error: some files/attrs were not transferred (see previous errors) (code 23) at main.c(1209) [sender=3.1.3]
+ res=23
+ '[' 23 = 24 ']'
+ '[' 23 = 23 ']'
+ '[' -n 1 ']'
+ res=0
+ exit 0
$
```

share/github-backup-utils/ghe-rsync
```diff
@@ -13,15 +13,30 @@ set -o pipefail
...
 # filter stderr while also keeping stdout and stderr separated.
-IGNOREOUT='^(file has vanished: |rsync warning: some files vanished before they could be transferred)'
-(rsync "${@}" $GHE_EXTRA_RSYNC_OPTS 3>&1 1>&2 2>&3 3>&- |
-  (egrep -v "$IGNOREOUT" || true)) 3>&1 1>&2 2>&3 3>&- |
-  (egrep -v "$IGNOREOUT" || true)
+ignoreout='^(file has vanished: |rsync warning: some files vanished before they could be transferred)'
+
+# Remove --ignore-missing-args parameter if using an earlier version of rsync.
+if ! rsync -h | grep -q '\-\-ignore-missing-args'; then
+  for parameter; do
```
lildude: Can you add the `in "$@"` part here too to make it clear this is what we're iterating over - it's not obviously clear unless you know about this behaviour.

share/github-backup-utils/ghe-rsync
snh: > That said, I'm not sure this is doing as it should be when run on hosts with support for this option:

I'm struggling to reproduce this, but will keep trying.

```
$ rsync -h | grep '\-\-ignore-missing-args'
     --ignore-missing-args   ignore missing source args without error

$ bash -x ./share/github-backup-utils/ghe-rsync --rsync-path=\'sudo -u git rsync\' --ignore-missing-args
[...]
+ ignoreout='^(file has vanished: |rsync warning: some files vanished before they could be transferred)'
+ rsync -h
+ grep -q '\-\-ignore-missing-args'
+ parameters=("$@")
+ egrep -v '^(file has vanished: |rsync warning: some files vanished before they could be transferred)'
+ rsync '--rsync-path='\''sudo' -u git 'rsync'\''' --ignore-missing-args
[...]
```

snh:
> I agree we need to allow the backup to continue though I fear we will be significantly increasing the perceived backup time with the additional verification step for all backups.

Some rudimentary testing with ~120,000 repositories indicated that the verification added around 5 minutes to the backup time (the majority of which was spent doing the `find`).

lildude:
> Some rudimentary testing with ~120,000 repositories indicated that the verification added around 5 minutes to the backup time (the majority of which was spent doing the find).

What is this as a proportion of the overall backup time? 

5 mins extra on a 15 min backup is significant. 5 mins extra on a 6 hour backup is barely noticeable.

lildude:
@snh nudge ☝️

snh:
@lildude Here are some timings from a different appliance, containing realistic end-user data:

```
$ find storage/ -mindepth 4 -maxdepth 4 -type f -print | sort | uniq | wc -l
129643
$ find repositories/ -mindepth 6 -maxdepth 7 -type d -name \*.git -exec dirname {} \; | sort | uniq | wc -l
14129

ghe-backup-repositories - Generating routes took 3s
ghe-backup-repositories - Fetching routes took 1s
ghe-backup-repositories - Processing routes took 0s
ghe-backup-repositories - Repo sync took 302s
ghe-backup-repositories - Special Data Directories Sync took 0s
ghe-backup-repositories - Verifying Routes took 15s
ghe-backup-repositories took 321s

ghe-backup-storage - Generating routes took 10s
ghe-backup-storage - Fetching routes took 3s
ghe-backup-storage - Processing routes took 1s
ghe-backup-storage - Storage object sync took 97s
ghe-backup-storage - Verifying Routes took 17s
ghe-backup-storage took 128s
```

avary33:
Thks

lildude:
Overall impact on performance for much larger instances seems quite small so I think we're good on that front. 

Think we should wrap the checks in an environment variable setting so it can be disabled easily if it does prove to have a big impact on other customers?

terrorobe:
Thinking out loud - do we have any good mechanism to signal these failures to the administrator, now that the backups are allowed to complete?

share/github-backup-utils/ghe-backup-repositories
```diff
@@ -362,4 +364,15 @@ RULES
 done
 bm_end "$(basename $0) - Special Data Directories Sync"
 
+if [ -z "$GHE_SKIP_ROUTE_VERIFICATION" ]; then
+  bm_start "$(basename $0) - Verifying Routes"
+
+  cat $tempdir/*.rsync | sort | uniq > $tempdir/source_routes
+  (cd $backup_dir/ && find * -mindepth 5 -maxdepth 6 -type d -name \*.git -exec dirname {} \; | sort | uniq) > $tempdir/destination_routes
+
+  git diff -- $tempdir/source_routes $tempdir/destination_routes || echo "Warning: One or more repository networks and/or gists were not found on the source appliance. Please contact GitHub Enterprise Support for assistance."
```
larsxschneider: @snh I suspect these warnings go usually unnoticed. Would it make sense to consider this a `failure` and process it accordingly? https://github.com/github/backup-utils/blob/0b42e114d2e7d47f81f565ae1d625f3d9098fdae/bin/ghe-backup#L229-L236

larsxschneider: TBH I don't understand that completely yet. Doesn't this check only ensure that all repo **directories** are present on the appliance and on the backup? If `rsync` fails with a 23 error and a single file is not transferred, then we wouldn't learn about it, right?

I saw a number of those errors in the wild:
```
Jun 04 21:39:30 rsync error: some files/attrs were not transferred (see previous errors) (code 23) at main.c(1505) [generator=3.0.6]
```

larsxschneider: Plus, wouldn't that check always generate a false warning if repositories have been deleted during the backup run? Do we delete repositories while the backup is running?

snh: > Would it make sense to consider this a failure and process it accordingly?

These errors indicate a consistency issue on the appliance rather than an issue with the backup itself, therefore the backup is still to be considered valid, and not a failure.

> TBH I don't understand that completely yet. Doesn't this check only ensure that all repo directories are present on the appliance and on the backup?

It is intended to check that all routes exist in the backup, it is not intended to detect individual file transfer errors from `rsync`, which should instead be detected through other means.

> Do we delete repositories while the backup is running?

We pause purging during the relevant stage of the backup. This is not an issue.

Issue #98, opened 2015-01-29
Backup to mounted samba share fails silently


juchiu:
bin/ghe-backup Lines 53-60

``` shell
# Mark the snapshot as in-progress by creating the symlink. If this fails, it
# means another ghe-backup run is already in progress and we should exit.
# NOTE: The -n argument to ln is non-POSIX but widely supported.
if ! ln -sn "$GHE_SNAPSHOT_TIMESTAMP" ../in-progress 2>/dev/null; then
    snapshot="$(readlink ../in-progress)" ### Does not get past this point
    echo "Error: backup of $GHE_HOSTNAME already in progress in snapshot $snapshot. Aborting." 1>&2
    exit 1
fi

```

If you try to have `$GHE_DATA_DIR` as a samba mount, `ln` fails to create symlinks. 
When the line with `readlink` fails, the script exits with exit code 1 without any output.

Perhaps this is just an issue with the CIFS driver, but it is very confusing when there is no output.

Efpophis:
This is an issue on any filesystem that doesn't support linux-style symbolic links. FWIW, it also prevents using the backup utilities in a git-bash shell on Windows, placing the backup data onto (for example) a pre-formatted USB drive, and a host of other things that could be potentially useful.

PR #388, opened 2018-04-03
Check filesystem supports hardlinks

lildude:
As backup-utils uses rsync's hardlinking functionality for efficient backup storage, we should really be checking and enforcing this requirement.  This PR introduces an early check on the filesystem and exits with an error.

Fixes https://github.com/github/backup-utils/issues/98 - though we don't use the method that caused that problem anymore but we do still use hardlinks.

### DIFF
--- a/bin/ghe-backup
+++ b/bin/ghe-backup
@@ -57,10 +57,10 @@ cd "$GHE_SNAPSHOT_DIR"
 touch "incomplete"
 
 # Exit early if the snapshot filesystem doesn't support hard links
-if ln $0 test_link >/dev/null 2>&1; then
+if ln incomplete test_link >/dev/null 2>&1; then
   rm -f test_link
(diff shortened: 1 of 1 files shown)


bin/ghe-backup
```diff
@@ -56,6 +56,14 @@ mkdir -p "$GHE_SNAPSHOT_DIR"
 cd "$GHE_SNAPSHOT_DIR"
 touch "incomplete"
 
+# Exit early if the snapshot filesystem doesn't support hard links
+if ln incomplete test_link >/dev/null 2>&1; then
+  rm -f test_link
+else
+  echo "Error: the filesystem containing $GHE_DATA_DIR does not support hard links." 1>&2
```
dbussink: @lildude Should this also explain why we need hard links?

PR #389, opened 2018-04-13
Switch back to optimised route restores

lildude:
When unifying the backup and restore process as part of https://github.com/github/backup-utils/pull/375, I unified all restores to use the same approach, but this happened to be the unoptimised of the two. This PR switches everything back to using the optimised method.

It also adds more output that can be used to debug the behaviour and performance.


01652242454:
tuyệt

PR #391, opened 2018-04-21
Optionally log verbose output to a file instead of stdout

jatoben:
Verbose logging is super helpful for debugging backup problems, but writing [all verbose output to stdout](https://github.com/github/backup-utils/blob/ff640e7972d4724511eeb1c9e7c2eb826c9564f1/share/github-backup-utils/ghe-backup-config#L29-L35) means that it can't easily be enabled if stdout should remain easily readable or mailable. For example, an admin might like to receive an email via cron after a backup has completed, with the text showing the backup steps and [confirming when the backup run finished](https://github.com/github/backup-utils/blob/ff640e7972d4724511eeb1c9e7c2eb826c9564f1/bin/ghe-backup#L220).

This PR adds a `GHE_VERBOSE_LOG` setting, which may be pointed to a file that the verbose output will be written to. If a backup fails, the log can then be used to investigate the problem further.

/cc @snh @terrorobe

### DIFF
--- a/backup.config-example
+++ b/backup.config-example
@@ -23,6 +23,11 @@ GHE_NUM_SNAPSHOTS=10
 #
 #GHE_RESTORE_HOST="github-standby.example.com"
 
+# When verbose output is enabled with `-v`, it's written to stdout by default. If
+# you'd prefer it to be written to a separate file, set this option.
+#
+# GHE_VERBOSE_LOG="/var/log/backup-verbose.log"
+
 # Any extra options passed to the SSH command.
 # In a single instance environment, nothing is required by default.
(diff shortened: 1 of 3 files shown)


backup.config-example
```diff
@@ -23,6 +23,11 @@ GHE_NUM_SNAPSHOTS=10
 #
 #GHE_RESTORE_HOST="github-standby.example.com"
 
+# When verbose output is enabled with `-v`, it's written to stdout by default. If
+# you'd prefer it to be written to a separate file, set this option.
+#
+# GHE_VERBOSE_LOG="/var/log/backup-verbose.log"
```
gnawhleinad: > # 

nit: 🔥 the space to be consistent with the other commented-out examples.

PR #392, opened 2018-04-24
Switch back to optimised backup route calculation

lildude:
As with the restore routes, I inadvertently switched to a sub-optimal route calculation method in https://github.com/github/backup-utils/pull/375.

This PR switches back to the optimised backup routes calculation method as https://github.com/github/backup-utils/pull/389 did for the restore routes.

### DIFF
--- a/share/github-backup-utils/ghe-backup-repositories
+++ b/share/github-backup-utils/ghe-backup-repositories
@@ -63,7 +63,7 @@ user="${host%@*}"
 
 hostnames=$host
 ssh_config_file_opt=
-tempdir=$(mktemp -d -t backup-utils-restore-XXXXXX)
+tempdir=$(mktemp -d -t backup-utils-backup-XXXXXX)
 ghe-ssh "$GHE_HOSTNAME" -- mkdir -p $tempdir
 routes_list=$tempdir/routes_list
 local_routes_list=$tempdir/local_routes_list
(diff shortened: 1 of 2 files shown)


share/github-backup-utils/ghe-backup-repositories
```diff
@@ -64,6 +64,8 @@ user="${host%@*}"
 hostnames=$host
 ssh_config_file_opt=
 tempdir=$(mktemp -d -t backup-utils-restore-XXXXXX)
```
snh: The `tempdir` needs to be created on the remote appliance too, as per https://github.com/github/backup-utils/blob/legacy/share/github-backup-utils/ghe-backup-alambic-cluster-ng#L40.

share/github-backup-utils/ghe-backup-storage
```diff
@@ -37,6 +37,8 @@ user="${host%@*}"
 hostnames=$host
 ssh_config_file_opt=
 tempdir=$(mktemp -d -t backup-utils-restore-XXXXXX)
```
snh: As mentioned, the `tempdir` needs to be created on the remote appliance too, as per https://github.com/github/backup-utils/blob/legacy/share/github-backup-utils/ghe-backup-alambic-cluster-ng#L40.

share/github-backup-utils/ghe-backup-repositories
lildude: Whoops 😊. Good catch. Fixed in https://github.com/github/backup-utils/pull/392/commits/7b3381aa4a1fb14a42a2d2ff1fab6f3d7dcedc14

PR #387, opened 2018-04-03
Improve code style, the styleguide and implement ShellCheck

lildude:
Whilst working on https://github.com/github/backup-utils/pull/375 I noticed quite a few inconsistencies and little niggles with the code that needed to be addressed but didn't warrant including in that PR. This PR addresses those niggles, namely:

- consistent indentation,
- conformance with the styleguide
- updating of the styleguide
- and implementing ShellCheck recommendations and a test for it too.

A lot of the changes in this PR are around whitespace so you probably wan't to use the `?w=1` whitespace "hack".

PR #394, opened 2018-04-24
Fix restoring of audit and hookshot logs to cluster nodes

lildude:
I've just discovered I made a syntactical error when implementing the Shellcheck recommendations in https://github.com/github/backup-utils/pull/387 that broke the restoring of hookshot and audit logs to a cluster: I put the `-print0` in the wrong place. 😊

This PR fixes that by moving the `-print0` to the end of the `find` command.

I've also added a test to catch this in future.

### DIFF
--- a/test/testlib.sh
+++ b/test/testlib.sh
@@ -252,13 +252,13 @@ setup_test_data () {
 
   mkdir -p "$loc/audit-log/"
   cd "$loc/audit-log/"
-  touch audit_log-1-$last_yr-$last_mth-1.gz
-  touch audit_log-1-$this_yr-$this_mth-1.gz
+  echo "fake audit log last yr last mth" | gzip > audit_log-1-$last_yr-$last_mth-1.gz
+  echo "fake audit log this yr this mth" | gzip > audit_log-1-$this_yr-$this_mth-1.gz
 
   # Create hookshot logs
   mkdir -p "$loc/hookshot/"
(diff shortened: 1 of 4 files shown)


test/testlib.sh
```diff
@@ -252,13 +252,13 @@ setup_test_data () {
...
-  touch audit_log-1-$last_yr-$last_mth-1.gz
-  touch audit_log-1-$this_yr-$this_mth-1.gz
+  echo "fake audit log last yr last mth" | gzip > audit_log-1-$last_yr-$last_mth-1.gz
+  echo "fake audit log this yr this mth" | gzip > audit_log-1-$this_yr-$this_mth-1.gz
 
   # Create hookshot logs
   mkdir -p "$loc/hookshot/"
   cd "$loc/hookshot/"
-  touch hookshot-logs-2018-03-05.gz
+  echo "fake hookshot log" | gzip > hookshot-logs-2018-03-05.gz
```
dctucker: Should this be a hard-coded date-based filename?

lildude: Sort of. The filename needs to be in the format: `<index_name>-<year>-<mth>-<day>.gz` as this is how we name the file when backing up, so the actual date isn't really relevant. It's simpler to use a hard-coded date then try get fancy with dynamically generating it.  We're only interested in the fact it's being processed in the test, regardless of its filename.

Issue #390, opened 2018-04-17
Restoring data taken from a node in HA to a standalone node creates an unnecessary storage host


doublemarket:
When restoring data taken from a node in an HA configuration to a standalone node, an unnecessary storage host which is a non-voting host in the source HA configuration is created on the standalone node.

**Example:**

A primary node in an HA configuration (backup source) has the following storage hosts:

```
source-primary$ ghe-spokes server show
+-------------------------------------------------+--------+-----------+-----------+------------+----+----------+--------+-------+-----+--------+
| Host                                            | Online | Embargoed | Quiescing | Evacuating | IP | Location | Voting | HDD   | Age | Uptime |
+-------------------------------------------------+--------+-----------+-----------+------------+----+----------+--------+-------+-----+--------+
| git-server-b97afa48-41df-11e8-acb7-0af065d42e64 | true   | false     | false     | false      |    | default  | true   | false |     | 0d     |
| git-server-d6fce9dc-41df-11e8-baec-0aed5a338dc8 | true   | false     | false     | false      |    | default  | false  | false |     | 0d     |
+-------------------------------------------------+--------+-----------+-----------+------------+----+----------+--------+-------+-----+--------+
```

The target standalone node has the following storage host before restoring:

```
target-standalone$ ghe-spokes server show
+-------------------------------------------------+--------+-----------+-----------+------------+----+----------+--------+-------+-----+--------+
| Host                                            | Online | Embargoed | Quiescing | Evacuating | IP | Location | Voting | HDD   | Age | Uptime |
+-------------------------------------------------+--------+-----------+-----------+------------+----+----------+--------+-------+-----+--------+
| git-server-31a081da-322e-11e8-85cb-0a65bdc21e62 | true   | false     | false     | false      |    |          | true   | false |     | 19d    |
+-------------------------------------------------+--------+-----------+-----------+------------+----+----------+--------+-------+-----+--------+
```

However, the hosts in the source appear after restoring data taken from the source:

```
target-standalone$ ghe-spokes server show
+-------------------------------------------------+--------+-----------+-----------+------------+----+----------+--------+-------+-----+--------+
| Host                                            | Online | Embargoed | Quiescing | Evacuating | IP | Location | Voting | HDD   | Age | Uptime |
+-------------------------------------------------+--------+-----------+-----------+------------+----+----------+--------+-------+-----+--------+
[...]

PR #396, opened 2018-04-26
Clean up stale HA nodes on restore

lildude:
Since the change in the method used to replicate repository, pages and storage data between nodes in a high availability (HA) configuration in GitHub Enterprise 2.10, restoring a backup of the HA node could result in "extra" nodes when re-establishing replication after restoring.

This PR addresses that by cleaning up and stale node entries immediately after restoring the backup.

Fixes https://github.com/github/backup-utils/issues/390


lildude:
Nudgy McNudgeface @github/backup-utils

snh:
@lildude Apologies for the delay here, I'll have a look at this tonight.

bin/ghe-restore
```diff
@@ -341,6 +341,16 @@ else
   fi
 fi
 
+# Clean up all stale replicas
+if ! $CLUSTER; then
+  inactive_nodes=$(echo "ghe-spokes server show inactive --json | jq -r '.[] | select(.host | contains(\"git-server\")).host' | sed 's/^git-server-//g'" | ghe-ssh "$GHE_HOSTNAME" -- /bin/bash)
```
snh: Unfortunately this fails to cleanup the stale HA nodes if the target instance is in an unconfigured state (and the replica(s) were online at the time of the backup).

This occurs because when unconfigured, we don't automatically run `ghe-config-apply` near the end of the restore, which results in the nodes still appearing as `online` in the database. It takes a `ghe-config-apply` for these stale nodes to transition to a `offline` state.

As an alternative, could we drop the `inactive` from the `ghe-spokes server show` and instead just cleanup any nodes that don't match the UUID that we restored (as found in `$GHE_RESTORE_SNAPSHOT_PATH/uuid`)?

lildude: Ooh, good point. I completely forgot about restoring to an unconfigured instance. Your suggestion makes sense too.

bin/ghe-restore
```diff
@@ -343,10 +343,14 @@ fi
 
 # Clean up all stale replicas
 if ! $CLUSTER; then
-  inactive_nodes=$(echo "ghe-spokes server show inactive --json | jq -r '.[] | select(.host | contains(\"git-server\")).host' | sed 's/^git-server-//g'" | ghe-ssh "$GHE_HOSTNAME" -- /bin/bash)
-  if [ -n "$inactive_nodes" ]; then
+  restored_uuid=$(cat $GHE_RESTORE_SNAPSHOT_PATH/uuid)
+  other_nodes=$(echo "ghe-spokes server show --json | jq -r '.[] | select(.host | contains(\"git-server\")).host' | sed 's/^git-server-//g'" | ghe-ssh "$GHE_HOSTNAME" -- /bin/bash)
```
snh: To reduce lines of code by 3, would we just add a `| grep -v $restored_uuid` to the end of the large collection of piped commands here? Or would this reduce the readability too much?

lildude:
@snh done, and we got tests too 😄

PR #399, opened 2018-05-02
Removing unneeded COPY from Dockerfile

elordahl:
Is this `COPY` to permit overriding functionality/convenience? The `ghe-docker-init` file is copied from the the previously downloaded `tar.gz`, so removing this line removes the need to have this repository cloned altogether.


elordahl:
Great!  Travis doesnt look happy, but i'm not sure the failure is related to this change (?)

PR #398, opened 2018-05-01
Noting backup-utils does not support replica backups

joshuajleigh:
Working with Github Enterprise Support I have learned that Github Backup Utilities does not currently support backing up a High Availability Replica instance. I wanted to be sure this was documented for others in the future.

PR #400, opened 2018-05-04
Block backing up replicas

lildude:
By the nature of high availability (HA) replication's transient state of some of the data at a point in time during synchronisation, and the fact the replica may lag behind the primary node without any adverse effects, GitHub has discouraged, but never enforced, backing up of HA environments via the replica.

This PR now enforces this and complements the documentation change coming in https://github.com/github/backup-utils/pull/398

This enforcement helps ensure that all the data backed up is current and reliable.


bin/ghe-host-check
```diff
@@ -91,6 +91,15 @@ if [ -z "$version" ]; then
   exit 2
 fi
 
+if [ -z "$GHE_ALLOW_REPLICA_BACKUP" ]; then
+  if [ -f "$GHE_REMOTE_ROOT_DIR/etc/github/repl-state" ] && [ "$(cat $GHE_REMOTE_ROOT_DIR/etc/github/repl-state)" = "replica" ]; then
```
lildude: Doh! This doesn't do what it should. Fixing 😊

lildude:
Silly mistake fixed 🤫

PR #423, opened 2018-08-07
Capture and display repository/gist warnings during cluster restore

jatoben:
The change in #386 allowed a backup to complete if a data consistency issue occured, printing a warning instead of causing a hard failure. Currently, an analogous situation exists for cluster restores; any errors along the way will cause the restore to fail. These may be minor issues that can be addressed with the help of GitHub support, and do not need to block the use of the destination cluster in the meantime.

This PR captures the stderr output of the repository and Gist finalize scripts from the cluster side, displaying them as warnings. Note that this will need a corresponding change on the Enterprise appliance side so that those scripts don't exit non-zero on error to take effect.

### DIFF
--- a/test/test-ghe-restore.sh
+++ b/test/test-ghe-restore.sh
@@ -467,7 +467,7 @@ begin_test "ghe-restore cluster"
   grep -q "dpages-cluster-restore-finalize OK" "$TRASHDIR/restore-out"
 
   # verify no warnings printed
-  grep -qv "Warning" "$TRASHDIR/restore-out"
+  ! grep -q "Warning" "$TRASHDIR/restore-out"
 
   # Verify all the data we've restored is as expected
   verify_all_restored_data


test/test-ghe-restore.sh
```diff
@@ -466,8 +466,57 @@ begin_test "ghe-restore cluster"
   grep -q "Pages to git-server-fake-uuid2" "$TRASHDIR/restore-out"
   grep -q "dpages-cluster-restore-finalize OK" "$TRASHDIR/restore-out"
 
+  # verify no warnings printed
+  grep -qv "Warning" "$TRASHDIR/restore-out"
```
lildude: This will always succeed if the `restore-out` file contains more than just "Warning" lines:

```console
$ cat foo.out
Warning
baaa
$ grep -qv Warning foo.out; echo $?
0
$
```

You would need to negate a successful find:

```console
$ grep -q Warning foo.out; echo $?
0
$ ! grep -q Warning foo.out; echo $?
1
$ ! grep -q foo foo.out; echo $?
0
$
```

jatoben: Ah, thanks for that @lildude 🙇 Fix coming up.

PR #442, opened 2018-10-01
consul recover from uuid change

deleonjavier:
`ghe-restore` was restoring a `uuid` file that consul now heavily relies on. 

/cc @lildude

### DIFF
--- a/bin/ghe-restore
+++ b/bin/ghe-restore
@@ -258,7 +258,9 @@ if [ -s "$GHE_RESTORE_SNAPSHOT_PATH/uuid" ] && ! $CLUSTER; then
   echo "Restoring UUID ..."
   cat "$GHE_RESTORE_SNAPSHOT_PATH/uuid" |
   ghe-ssh "$GHE_HOSTNAME" -- "sudo sponge '$GHE_REMOTE_DATA_USER_DIR/common/uuid' 2>/dev/null"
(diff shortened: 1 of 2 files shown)


bin/ghe-restore
```diff
@@ -258,7 +258,9 @@ if [ -s "$GHE_RESTORE_SNAPSHOT_PATH/uuid" ] && ! $CLUSTER; then
   echo "Restoring UUID ..."
   cat "$GHE_RESTORE_SNAPSHOT_PATH/uuid" |
   ghe-ssh "$GHE_HOSTNAME" -- "sudo sponge '$GHE_REMOTE_DATA_USER_DIR/common/uuid' 2>/dev/null"
-fi
+  ghe-ssh "$GHE_HOSTNAME" -- "sudo systemctl stop consul"
```
lildude: You'll need to create a dummy `systemctl` command in `test/bin`. A symlink to `ghe-fake-true` should do the trick as we don't care about the results of this cmd when testing.

sbryant: yeah, can probably just `|| true` this

deleonjavier: https://github.com/github/backup-utils/pull/442/commits/335c88a99fd7ccf85f5759485c24840d84ab162b

PR #456, opened 2018-11-16
[WIP] Attempt parallel restore

djdefi:
Attempt at parallelising where were are currently doing a serial operation for cluster restores.

cc: @lildude


share/github-backup-utils/ghe-restore-repositories
```diff
@@ -152,8 +152,13 @@ for file_list in $tempdir/*.rsync; do
     --rsync-path="sudo -u git rsync" \
     --files-from=$file_list \
     "$GHE_DATA_DIR/$GHE_RESTORE_SNAPSHOT/repositories/./" \
-    "$server:$GHE_REMOTE_DATA_USER_DIR/repositories/" 1>&3
+    "$server:$GHE_REMOTE_DATA_USER_DIR/repositories/" 1>&3 2>&3 &
 done
+
+for pid in $(jobs -p); do
+  wait "$pid"
```
terrorobe: I'm wondering if we could get away with a plain `wait` here.

`jobs` may IIRC include other backgrounded jobs started before `ghe-backup`.

From the `bash` manpage:

```
       wait [-n] [n ...]
              Wait for each specified child process and return its termination
              status.  Each n may be a process ID or a job specification; if a
              job spec is given, all processes  in  that  job's  pipeline  are
              waited  for.  If n is not given, all currently active child pro‐
              cesses are waited for, and the return status is zero.  If the -n
              option  is  supplied,  wait  waits  for any job to terminate and
              returns its exit status.  If n specifies a non-existent  process
              or  job, the return status is 127.  Otherwise, the return status
              is the exit status of the last process or job waited for.
```

Waiting for the current child processes sounds like what we want to do here

djdefi: I'm not familiar with using that, feel free to propose a change or commit to my branch directly. 

It looks like we are already seeing worse performance in testing, so we may need to re-evaluate this overall.

snh: From memory we do it this way to pick up any failing jobs. A plain `wait` on its own only returns the exit code of the last job to finish. In the past this has caused us to not detect failing jobs.

djdefi:
closing as stale for now

PR #472, opened 2019-02-18
Detect storage user on each host being backed up or restored

jatoben:
If a backup or restore targets a GitHub Enterprise Server cluster node that was upgraded from a previous release to the 2.16 series and isn't running the storage service, the user account detection added in https://github.com/github/backup-utils/pull/448 will return an incorrect result. `rsync` then won't have the correct permissions to copy storage data.

This PR aims to fix that by running the user detection on each storage node, rather than the target of the `ghe-backup` or `ghe-restore`.

### DIFF
--- a/share/github-backup-utils/ghe-backup-storage
+++ b/share/github-backup-utils/ghe-backup-storage
@@ -113,12 +113,11 @@ if ! ls $tempdir/*.rsync >/dev/null 2>&1; then
   exit 0
 fi
 
-storage_user=$(ghe-ssh "$GHE_HOSTNAME" -- stat -c %U /data/user/storage || echo git)
-
 # rsync all the storage objects
 bm_start "$(basename $0) - Storage object sync"
 for file_list in $tempdir/*.rsync; do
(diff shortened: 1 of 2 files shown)


jchamber904:
@jatoben This problem has been causing our backups to failed since we upgraded to 2.16.1 over the weekend.  Getting Access Denied error messages during the backup when accessing the storage nodes.  When SSH'ing to the storage nodes, we noticed ownership changed from user git to user alambic.

jchamber904:
@jatoben We are currently pulling from:
```
git clone -b stable --depth 1 https://github.com/github/backup-utils.git "$WORKSPACE/backup-utils"
```
Do we continue to pull from this location to get this patch?  When will this be merged?

jatoben:
@jchamber904 I'll merge this now, but it won't be added to the _stable_ branch until the next release is made. In the meantime, I believe support have shared a workaround that will avoid the problem. If you have any questions about that, please follow up in a support ticket.

Thanks!

PR #473, opened 2019-02-20
add Dockerfile based on alpine

JordanSussman:
This PR adds a Dockerfile based on alpine instead of Debian.

Why the need for something besides Debian? My organization typically blocks any images that have known high/critical CVE, and Debian images have numerous.

The vast majority of the Dockerfile matches that of the orginal, but has a few modifications that include:

- Install [gawk](https://pkgs.alpinelinux.org/package/edge/main/x86/gawk) to override the busybox version of awk
- Install [procps](https://pkgs.alpinelinux.org/package/v3.4/main/x86/procps) to overide the busybox version of ps
- Install openssh package instead of ssh

Both of the afermentioned busybox changes are needed to make those utilies respond in the same fashion as the debian equivalents

### DIFF
--- a/Dockerfile.alpine
+++ b/Dockerfile.alpine
@@ -1,0 +1,22 @@
+FROM alpine:latest
+
+RUN apk --update --no-cache add \
+    tar \
+    rsync \
+    ca-certificates \
+    openssh \
+    git \
+    bash \
+    gawk \
+    procps
+
+WORKDIR /backup-utils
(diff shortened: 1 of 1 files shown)


oakeyc:
Hey @JordanSussman, this PR is a bit old, but if you still want it, you can merge it :)

JordanSussman:
Hi @oakeyc, feel free to merge this.

PR #476, opened 2019-03-01
Restore target is ignored when specified on the command line

taz:
The change places the command line restore host arg into a temporary holding variable (`GHE_RESTORE_HOST_OPT`) so it's not stomped on by the version imported from the `backup.config` file.

It then either copies the target into `GHE_HOSTNAME` after the config file has been loaded or uses `GHE_RESTORE_HOST` if a target wasn't specified on the command line.

I believe I introduced this bug in https://github.com/github/backup-utils/pull/449/files 😞.

However, I'm unsure why the test at https://github.com/github/backup-utils/blob/master/test/test-ghe-restore.sh#L191-L213 didn't pick it up? I'm still investigating that, but I wanted to surface this bug now.

/cc @lildude - FYI and 👀on the test if you get a chance before I revisit it again next week.
/cc @jatoben - As we'd discussed the problem previously.

### DIFF
--- a/bin/ghe-restore
+++ b/bin/ghe-restore
@@ -63,7 +63,7 @@ while true; do
       ;;
     *)
       if [ -n "$1" ]; then
-        GHE_RESTORE_HOST="$1"
+        GHE_RESTORE_HOST_OPT="$1"
         shift
       else
         break
@@ -86,7 +86,7 @@ cleanup () {
 . "$( dirname "${BASH_SOURCE[0]}" )/../share/github-backup-utils/ghe-backup-config"
 
 # Grab the host arg
-GHE_HOSTNAME="$GHE_RESTORE_HOST"
+GHE_HOSTNAME="${GHE_RESTORE_HOST_OPT:-$GHE_RESTORE_HOST}"
 
 # Hostname without any port suffix
 hostname=$(echo "$GHE_HOSTNAME" | cut -f 1 -d :)
--- a/test/test-ghe-restore.sh
+++ b/test/test-ghe-restore.sh
@@ -188,6 +188,40 @@ begin_test "ghe-restore into unconfigured vm"
 )
 end_test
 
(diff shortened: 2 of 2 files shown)


lildude:
> However, I'm unsure why the test at https://github.com/github/backup-utils/blob/master/test/test-ghe-restore.sh#L191-L213 didn't pick it up?

🤔 I'm not sure either from a quick look. I'd need to experiment to see the output from the restore on this test but I suspect the test is finding a `localhost` in the output that may not be directly related to the restore itself.

taz:
I suspect you're right, I'll run through it again soon and see if I can get to the bottom of it.

madriga1:
Looking forward to this fix. We ran into this yesterday and it took a while to sort it out.

taz:
@lildude I think I've isolated the test failure. It appears that `ghe-restore` works as expected when a `GHE_RESTORE_HOST` target is added via an environment variable. However when added via the `backup.config` file the command line target is ignored.

So the original test was working, however only for one particular scenario.

I've tweaked the existing test to make the failure more obvious. If the command line argument isn't used, it will result in a connection failure now rather than just a mismatch: https://github.com/github/backup-utils/pull/476/files#diff-f8b59fe8478272b3950e85521fdb636fR235

I've also added an additional test which creates a temporary `backup.config` file and inserts a `GHE_BACKUP_HOST` attribute into it: https://github.com/github/backup-utils/pull/476/files#diff-f8b59fe8478272b3950e85521fdb636fR191-R223

To test it I've branched off `master` and added only the new tests at  https://github.com/github/backup-utils/compare/taz/test-restore-target-fail, which fails as expected.

The same tests have been included in this branch now as well and pass, so I believe they're more accurate. I wasn't sure the best practice for introducing the modified `backup.config` file though, what do you think?

taz:
@lildude would you mind casting an eye over the updated tests when you get a chance? 🙇

PR #477, opened 2019-03-11
Make audit log backup process aware of MySQL import

juruen:
We need to make `backup utils` aware of the audit log import to MySQL.

The state of the migration, or to be more precise, whether or not the migration is complete is marked by a flag in `/data/user/common/audit-log-import/complete`.

This the proposal to handle this:

- At backup time
  1. If the flag exists, we don't need to do anything else apart from making sure the flag itself makes it to the snapshot
  2. If the flag doesn't exist, we back up the ES indices as usual

- At restore time
  1. If the flag exists in the snapshot, we don't need to do anything else apart from make sure we create the flag in the instance's filesystem
  2. If the flag doesn't exist, we restore the ES indices as usual

Some implementation details:

- I removed the code to check an old transition that is not needed any longer
- I basically wrapped the old method into their respective functions and added logic to call them when the flag is not present.

### DIFF
--- a/share/github-backup-utils/ghe-backup-es-audit-log
+++ b/share/github-backup-utils/ghe-backup-es-audit-log
@@ -10,6 +10,34 @@ set -e
 # shellcheck source=share/github-backup-utils/ghe-backup-config
 . "$( dirname "${BASH_SOURCE[0]}" )/ghe-backup-config"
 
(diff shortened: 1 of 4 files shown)


juruen:
Closing this in favour of #485

PR #478, opened 2019-03-12
Adding a new option to create a staging instance safely

lowply:
[Setting up a staging instance - GitHub Help](https://help.github.com/en/enterprise/2.16/admin/installation/setting-up-a-staging-instance) says that the recommended way to create a staging instance is to restore the latest backup data to a fresh instance.

However, the `ghe-restore` command restores the */data/user/common/uuid* file and it causes some problems especially when the staging instance is in the same network as the production instance. For example, I saw a case where a staging HA pair failed to restart replication after upgrading to GitHub Enterprise 2.16 series. Turns out it happened because Consul failed to choose the appropriate leader node due to the the duplicate UUIDs in the same network.

With this pull request, I'd like to suggest that we add the `--staging` option to the `ghe-restore` command so that the command restores everything except the uuid file.

### DIFF
--- a/bin/ghe-restore
+++ b/bin/ghe-restore
@@ -17,6 +17,8 @@
 #/   -v | --verbose    Enable verbose output.
 #/   -h | --help       Show this message.
 #/        --version    Display version information and exit.
+#/        --staging    Use this option to create a staging instance. Restore
+#/                     everything except UUID.
 #/   -s <snapshot-id>  Restore from the snapshot with the given id. Available
 #/                     snapshots may be listed under the data directory.
 #/   <host>            The <host> is the hostname or IP of the GitHub Enterprise
@@ -30,6 +32,7 @@ set -e
 
 # Parse arguments
 restore_settings=false
+restore_staging=false
 force=false
 while true; do
   case "$1" in
@@ -43,6 +46,10 @@ while true; do
       ;;
     -c|--config)
       restore_settings=true
+      shift
+      ;;
+    --staging)
+      restore_staging=true
       shift
       ;;
     -h|--help)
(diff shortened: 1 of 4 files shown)


lowply:
Hi @github/pe-build-reviewers 👋 What do you think about this change?

dbussink:
> Hi @github/pe-build-reviewers 👋 What do you think about this change?

@lowply If the `uuid` file is not restored, it will reset it with a new `uuid`, but that also means that none of the repositories on the restored machine will work, as routing information in the MySQL database still points to the old `uuid`. Have you tested what happens in this case?

lowply:
@dbussink Thanks for the comment. I hadn't tested that, so I did it. You're right, if the `uuid` in the */data/user/common/uuid* and the database differ, I got the `500` error when browsing to the repository on the restored instance. It was the `GitHub::DGit::UnroutedError` exception.

To avoid this, I'd like to suggest that we `sed` the `mysql.sql` file to replace the backed up `uuid` with the target host's `uuid` right before the `ghe-import-mysql`. Here's the key change:

https://github.com/github/backup-utils/blob/721507d4aed0645931f4acae7dfaed34a4d2436d/share/github-backup-utils/ghe-restore-mysql#L41-L48

I added the `sed` part before the `ghe-import-mysql` if the `RESTORE_STAGING` flag is `true`.

I've tested this on my 2.16.5 instance and it worked. What do you think?

dbussink:
> I've tested this on my 2.16.5 instance and it worked. What do you think?

I'm not really sure if this is a really solid and resilient approach. There have also been discussions around using the clustering restore paths for single nodes. Can't we solve the issues around what happens if there's multiple instances with the same uuid on the network? Can those issues be prevented? 

I'm not really the person to make a call on this though, that's probably either in @github/pe-build or @github/pe-releases area then.

lowply:
> I'm not really sure if this is a really solid and resilient approach.

I agree with you, and I couldn't think of better way to solve this problem. For now I'm marking this PR as closed. Thank you for your comment @dbussink 😄

PR #480, opened 2019-03-28
Adds a minimum required sed version to the documentation

craigerrington:
The [`-E` flag](https://github.com/github/backup-utils/blob/cb78f08e7c2e26d36cc2f584e22b6e6bd37a426b/bin/ghe-host-check#L87) became available in sed 4.3 and above.

There are still some extended support distros (SLES 11 at least) which are shipping with older versions, so I think it would be good to call out the minimum requirement.

### DIFF
--- a/docs/requirements.md
+++ b/docs/requirements.md
@@ -6,7 +6,7 @@ storage and must have network connectivity with the GitHub Enterprise Server app
 ## Backup host requirements
 
 Backup host software requirements are modest: Linux or other modern Unix operating
-system with [bash][1], [git][2], [OpenSSH][3] 5.6 or newer, and [rsync][4] v2.6.4 or newer.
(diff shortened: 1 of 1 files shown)


aferreira:
I wonder if it would be better to downgrade the code that requires sed 4.3. That is, from:
```sh
version=$(echo "$output" | sed -E -n 's/GitHub Enterprise( Server)? version (.*)/\2/p')
```
to
```sh
version=$(echo "$output" | sed -n 's/GitHub Enterprise\( Server\)\{0,1\} version \(.*\)/\2/p')
```

In this case, it would not be necessary to document the `sed` required version and the code would work again for systems with older versions.

I am not 100% sure of how portable the second code is. My first try used `\( Server\)\?` but, even though that seems to work in Linux, it does not on OS X. That is why I went with the byzantine `\( Server\)\{0,1\}` that worked on both.

lildude:
We've gone the route of using a simplified `sed` in https://github.com/github/backup-utils/pull/509. Closing.

PR #481, opened 2019-03-28
ghe-host-check -F configfile

aferreira:
This is meant to check against the hostname of a config file
that was not put into place yet. Works like

    GHE_BACKUP_CONFIG=configfile ghe-host-check

but with a documented -F option.

One use case is validating a newly generated `backup.config` file
before putting it into production (for example, via Puppet).

### DIFF
--- a/bin/ghe-host-check
+++ b/bin/ghe-host-check
@@ -6,6 +6,7 @@
 #/ OPTIONS:
 #/   -h | --help       Show this message.
 #/        --version    Display version information.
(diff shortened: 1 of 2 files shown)


manue1:
Sorry for the delay on this one, I'm going to close this out since the desired usage of the tool can already be achieved today with the outlined command:
```
GHE_BACKUP_CONFIG=backup.config.X ghe-host-check
```

PR #487, opened 2019-05-01
Fix audit log import to MySQL flag removal for old snapshots and skip rsync'ed indices

juruen:
This PR solves two issues that could potentially cause data loss in the upcoming 2.17 version.

### Problem
If an old snapshot (<2.17) is restored in a 2.17 instance where the audit import to MySQL is marked as complete in `/data/user/common/audit-log-import/complete`, the restored instance wrongly keeps the flag that indicates the import is complete.

This could cause a data loss because under certain circumstances a new backup would assume that data is migrated to MySQL when it's not.

### Solution
The solution is simple: if a restored snapshot doesn't contain an `audit-log-mysql` directory, it means it's an old snapshot. In that case, we remove the flag on the instance where the restore is happening.

### Problem
Audit log indices can be backed up from:

- MySQL
- Elasticsearch through JSON dumps
- Elasticsearch through rsync

When `GHE_BACKUP_ES_AUDIT_LOGS=no` only MySQL should be used.

JSON dumps are correctly skipped, however all the indices including the audit log ones are rsynced in a subsequent step.

### Solution
When `GHE_BACKUP_ES_AUDIT_LOGS=no` and the import is complete, we configure the rsync step to ignore all the audit log indices.

To do so, we build a file that contains the  `UUIDs` of the audit log indices. Note this is what is used in disk and not the actual index name.

The UUIDs are fetched through a simple:

```
curl -s http://localhost:9201/_cat/indices/audit_log?h=uuid
```

The file containing this list  is passed as a`--exclude-from` rsync parameter that contains one directory/pattern to exclude per line.

### DIFF
--- a/share/github-backup-utils/ghe-backup-es-rsync
+++ b/share/github-backup-utils/ghe-backup-es-rsync
@@ -35,7 +35,7 @@ echo elasticsearch.yml >"$exclude_file"
 # as those indices will be rebuilt from MySQL during a restore
 if [ "$GHE_BACKUP_ES_AUDIT_LOGS" = "no" ] && ghe-ssh "$host" test -e "/data/user/common/audit-log-import/complete"; then
   ghe_verbose "* Excluding Audit Log indices"
-  ghe-ssh "$host" curl -s 'localhost:9201/_cat/indices/audit_log?h=uuid' >>$exclude_file 2>&3
+  ghe-ssh "$host" curl -s 'http://localhost:9201/_cat/indices/audit_log?h=uuid' >>$exclude_file 2>&3
 fi
 
 # Verify that the /data/elasticsearch directory exists.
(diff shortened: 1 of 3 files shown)


share/github-backup-utils/ghe-backup-es-rsync
```diff
@@ -27,6 +27,17 @@ fi
...
 
+# Create exclude file
+exclude_file="$(mktemp)"
+echo elasticsearch.yml >"$exclude_file"
+
+# Exclude audit log indices when configuration says so and import to MySQL is complete
+# as those indices will be rebuilt from MySQL during a restore
+if [ "$GHE_BACKUP_ES_AUDIT_LOGS" = "no" ] && ghe-ssh "$host" test -e "/data/user/common/audit-log-import/complete"; then
+  ghe_verbose "* Excluding Audit Log indices"
+  ghe-ssh "$host" curl -s 'localhost:9201/_cat/indices/audit_log?h=uuid' >>$exclude_file 2>&3
```
dctucker: Nit: is this curling an http endpoint? an `http://` would make this more obvious.

juruen: Certainly!

juruen: Fixed!

PR #494, opened 2019-05-28
add option to generate a persistent rsync log

larsxschneider:
Generate a log file with all rsync operations during a ghe-backup run.
This allows users to scan changed content between two consecutive
backup runs.

### DIFF
--- a/backup.config-example
+++ b/backup.config-example
@@ -28,6 +28,12 @@ GHE_NUM_SNAPSHOTS=10
 #
 #GHE_VERBOSE_LOG="/var/log/backup-verbose.log"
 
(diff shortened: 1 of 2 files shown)


snh:
When backing up or restoring a GitHub Enterprise Server Cluster environment, we have multiple copies of `rsync` running at the same time at numerous points in the backup and restore processes.

Will each of these writing to the same log file simultaneously cause any issues?

Also, how large can we expect these logs files to be for a typical environment?

larsxschneider:
I'll use `GHE_EXTRA_RSYNC_OPTS` instead of a custom option. Thanks @snh !

PR #500, opened 2019-06-12
check dependency versions and warn if they are outdated

larsxschneider:
I picked the checked versions somewhat arbitrarily based on the versions
shipped with Debian 9 (Stretch). The intention is to encourage users to
update. If these versions are not met, then everything works as before
as we only print a warning if the versions are not met.

Please note: This commit introduces Perl as dependency. However, this
should not be a problem as Git depends on Perl as well. Since Git is
required by backup-utils, we can assume that Perl is installed on the
backup host.

### DIFF
--- a/bin/ghe-backup
+++ b/bin/ghe-backup
@@ -41,6 +41,14 @@ done
 # shellcheck source=share/github-backup-utils/ghe-backup-config
 . "$( dirname "${BASH_SOURCE[0]}" )/../share/github-backup-utils/ghe-backup-config"
 
+# Recommend recent versions of the dependencies
+# Please keep these versions in sync with the docs!
+# https://github.com/github/backup-utils/blob/master/docs/requirements.md#backup-host-requirements
+check_version bash  --version 4.4.12
+check_version git   --version 2.11.0
+check_version rsync --version 3.1.2
+check_version ssh   -V        7.4
+
 # Used to record failed backup steps
 failures=
 
--- a/docs/requirements.md
+++ b/docs/requirements.md
@@ -6,7 +6,7 @@ storage and must have network connectivity with the GitHub Enterprise Server app
 ## Backup host requirements
 
 Backup host software requirements are modest: Linux or other modern Unix operating
-system with [bash][1], [git][2], [OpenSSH][3] 5.6 or newer, and [rsync][4] v2.6.4 or newer.
+system with [bash][1], [git][2], [OpenSSH][3] 7.4 or newer, and [rsync][4] v3.1.2 or newer.
 
 We encourage the use of [Docker](docker.md) if your backup host doesn't meet these
 requirements, or if Docker is your preferred platform.
--- a/share/github-backup-utils/ghe-backup-config
+++ b/share/github-backup-utils/ghe-backup-config
@@ -215,6 +215,22 @@ if ! type readlink 1>/dev/null 2>&1; then
   }
 fi
 
+# Check the version of a given command and print a warning if the minimum
+# recommended version is not met.
+check_version () {
+  local cmd=$1
+  local cmd_version_option=$2
+  local min_version=$3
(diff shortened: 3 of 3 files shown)


docs/requirements.md
```diff
@@ -6,7 +6,7 @@ storage and must have network connectivity with the GitHub Enterprise Server app
 ## Backup host requirements
 
 Backup host software requirements are modest: Linux or other modern Unix operating
-system with [bash][1], [git][2], [OpenSSH][3] 5.6 or newer, and [rsync][4] v2.6.4 or newer.
+system with [bash][1], [git][2], [OpenSSH][3] 7.4 or newer, and [rsync][4] v3.1.2 or newer.
```
lildude: I think you're going to need to explicitly call out Perl now as you are directly using it.

lildude: Additionally, other than encouraging a user to use later versions, is there any specific reason for these versions? My fear is seeing new version requirements is going to prompt support tickets or issues asking why.

I suspect customers using RHEL (and there are quite a lot) are most likely to suddenly start seeing these version messages as they tend to be using older versions of each of these pkgs.

bin/ghe-backup
```diff
@@ -41,6 +41,14 @@ done
 # shellcheck source=share/github-backup-utils/ghe-backup-config
 . "$( dirname "${BASH_SOURCE[0]}" )/../share/github-backup-utils/ghe-backup-config"
 
+# Recommend recent versions of the dependencies
+# Please keep these versions in sync with the docs!
+# https://github.com/github/backup-utils/blob/master/docs/requirements.md#backup-host-requirements
+check_version bash  --version 4.4.12
+check_version git   --version 2.11.0
+check_version rsync --version 3.1.2
+check_version ssh   -V        7.4
```
lildude: If this is not a hard requirement, we should hide these unless using verbose output.

docs/requirements.md
larsxschneider: The main reason for all this is to ensure `--ignore-missing-args` is available in `rsync`.  The flag was introduced in [`rsync` version `3.1.0`](https://download.samba.org/pub/rsync/src/rsync-3.1.0-NEWS).

If the flag is not available, `ghe-backup` [suppresses exits with code 23](
https://github.com/github/backup-utils/blob/5ea6eb3d87474a162f2f8982681a63ce77abae8c/share/github-backup-utils/ghe-rsync#L38-L41). That can mask potentially fatal `rsync` errors.

---

Do you think it would make sense to use RHEL 7 or something as base for the minimum versions except `rsync`?

lildude: > The main reason for all this is to ensure `--ignore-missing-args` is available in `rsync`. The flag was introduced in [`rsync` version `3.1.0`](https://download.samba.org/pub/rsync/src/rsync-3.1.0-NEWS).
> 
> If the flag is not available, `ghe-backup` [suppresses exits with code 23](https://github.com/github/backup-utils/blob/5ea6eb3d87474a162f2f8982681a63ce77abae8c/share/github-backup-utils/ghe-rsync#L38-L41). That can mask potentially fatal `rsync` errors.

🤔 I recall that change being introduced but I don't recall any scenarios of legit fatal errors this would suppress, which is why I think we went for that solution.

What fatal rsync errors is this suppressing?

If these really are fatal errors, then bumping rsync's _minimum_ requirement makes sense. If there's no specific reason for the others, I don't think we should change their requirements.

> Do you think it would make sense to use RHEL 7 or something as base for the minimum versions except `rsync`?

No. I think it should be the minimum versions required for the functionality we require, regardless of the OS revision. We deliberately don't make any OS recommendations because we don't and can't test against all of them.

lildude: Additionally, if we're going to bump the minimum `rsync` version because of the suppression of legitimate fatal errors, I think we should make this a hard requirement and remove all the workarounds for suppressing exit code 23 too.

PR #509, opened 2019-08-16
Replace "sed -E" in ghe-host-check with a more portable solution

taz:
On older backup hosts, `sed` doesn't include the `-E` option which results in a `ghe-host-check` failure. While replacing `sed` with a newer option would be ideal, it's not always achievable in a short timeframe.

The proposed solution requires the version string to be at the end of the first string returned—is that a valid long term assumption? It will cope with any future product name changes more gracefully however, which is what `sed -E` was added to accomodate.

### DIFF
--- a/bin/ghe-host-check
+++ b/bin/ghe-host-check
@@ -84,7 +84,7 @@ if [ $rc -ne 0 ]; then
   exit $rc
 fi
 
-version=$(echo "$output" | head -1 | awk '{print $NF}')
+version=$(echo "$output" | grep "GitHub Enterprise" | awk '{print $NF}')
 
 if [ -z "$version" ]; then
   echo "Error: failed to parse version on '$host' or this isn't a GitHub appliance." 1>&2
--- a/bin/ghe-host-check
+++ b/bin/ghe-host-check
@@ -84,7 +84,7 @@ if [ $rc -ne 0 ]; then
   exit $rc
 fi
 
-version=$(echo "$output" | sed -E -n 's/GitHub Enterprise( Server)? version (.*)/\2/p')
+version=$(echo "$output" | head -1 | awk '{print $NF}')
 
 if [ -z "$version" ]; then
   echo "Error: failed to parse version on '$host' or this isn't a GitHub appliance." 1>&2


bin/ghe-host-check
```diff
@@ -84,7 +84,7 @@ if [ $rc -ne 0 ]; then
   exit $rc
 fi
 
-version=$(echo "$output" | sed -E -n 's/GitHub Enterprise( Server)? version (.*)/\2/p')
+version=$(echo "$output" | head -1 | awk '{print $NF}')
```
snh: This can output various error messages that would get incorrectly matched here. An alternative approach would be:

```suggestion
version=$(echo "$output" | grep "GitHub Enterprise\( Server\)\? version" | awk '{print $NF}')
```

taz: Good catch there @snh, it didn't turn up in my earlier testing. It looks like the test suite is cranky too. I'll adjust, re-test, and then see if the tests are still failing.

lildude: 🤔 I've not tested, but is there a risk this `grep` is moving the assumption of using a modern variant of `sed` to `grep`?  I'm thinking some old versions of `grep` may behave like `fgrep` so can only use fixed strings.

An alternate solution is the one posed at https://github.com/github/backup-utils/pull/480#issuecomment-477835228 (also not tested)

```
version=$(echo "$output" | sed -n 's/GitHub Enterprise\( Server\)\{0,1\} version \(.*\)/\2/p')
```

taz: I tested the alternative solution above on MacOS and it works as expected, but I can't vouch for its portability either—I can probably find out though if we'd prefer to go down that path. 

However are we backing ourselves into a corner re: options here by:

1. Trying to do the check in one operation; and 
2. Looking for the word "Server"? 

Can we assume that as long as "GitHub Enterprise" is present, it doesn't matter if "Server" is also there?

Why not pull the "GitHub Enterprise" string out first, and then manipulate the result to get the version number. eg:

```
echo "$output" | grep "GitHub Enterprise" | awk '{print $NF}'
```

That's a fixed string grep followed by an `awk` variable that I'm pretty sure has been there for a long time...

lildude: Yeah, I think that would be safe to do.

PR #515, opened 2019-10-01
Update getting-started.md

sennap:
As someone who went through this process for the first time, I wanted to make note of some things in the guide that I was confused by, and propose some wording that might help save others from questions:

- Added note about creating a new backup.config file + example command to do that =
- Added ./ for running those ghe commands

cc @skhalife @glennwester @nielspineda

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -11,8 +11,11 @@
     **Note**: you will need to use [Backup Utilities v2.11.x][2] or the `legacy` branch to
     backup and restore GitHub Enterprise Server 2.10 and earlier.
 
- 2. Copy the [`backup.config-example`][3] file to `backup.config` and modify as
-    necessary. The `GHE_HOSTNAME` value must be set to the primary GitHub Enterprise Server
+ 2. Copy the [`backup.config-example`][3] file to a new `backup.config` file and modify as
+    necessary.
+    - In the backup-utils directory, run `cp backup.config-example backup.config`
+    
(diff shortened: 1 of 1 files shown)


docs/getting-started.md
```diff
@@ -31,13 +34,13 @@
     * In a clustering environment, the `GHE_EXTRA_SSH_OPTS` key must be configured
       with the `-i <abs path to private key>` SSH option.
 
- 3. Add the backup host's SSH key to the GitHub appliance as an *Authorized SSH
+ 3. Add the backup host's (in this case your computer) SSH key to the GitHub appliance as an *Authorized SSH
```
lildude: I'm not sure I understand this addition. To me, this addition suggests the SSH key the admin user uses to connect to the backup host is the one that needs to be added to the appliance, which isn't necessarily correct.

We encourage customers to use a dedicated host for their backups and if they're using good practices, the backup user will have their own SSH key which won't be the same as the one the admin uses.

Could you explain this addition a bit more so I can understand the background behind its addition.

sennap: hey @lildude 👋 ! sorry I completely forgot about this PR 🙈

I'm having a hard time remembering what my logic was there and I agree with what you're saying. It doesn't really make sense so I'm removing this line!

oakeyc:
@sennap would you still like this PR?

geeebse (approved):

geeebse (approved):

bwestover:
closing as stale

PR #516, opened 2019-10-02
Remove temporary exclude file from the backup host not the target GHES appliance

taz:
A temporary exclusion file is created on the backup host during the rsync of ES indexes in order to prevent audit logs from being unnecessarily transferred. 

During the post-script cleanup we mistakenly attempt to remove the temporary file from the target GHES server rather than the backup host. The result is that the backup host accumulates stale exclusion files over time.

The one line change here removes the `ghe-ssh` call from the front of the `rm` so the temporary file is removed as intended.


juruen:
@taz thank you for catching and fixing this! ❤️

PR #517, opened 2019-10-10
Prepare for audit log MySQL backend deprecation

juruen:
This PR makes changes to support the audit log MySQL backend deprecation.

There is a companion PR that has an extensive test suite in the GHES  repository.

For backwards compatibility, we need to make sure if we restore a snapshot that only contains MySQL entries, these entries need to be reindexed into Elasticsearch.

We also need to use flags to prevent two scenarios:

- Useless transitions to MySQL in 2.17 and 2.18
- Avoid the audit log table truncation by migrations while a reindex is necessary.

### DIFF
--- a/share/github-backup-utils/ghe-restore-audit-log
+++ b/share/github-backup-utils/ghe-restore-audit-log
@@ -45,10 +45,10 @@ set_remote_flag(){
   local msg=$2
 
   ghe_verbose "$2"
-  ghe-ssh "$GHE_HOSTNAME" -- "sudo touch $GHE_REMOTE_ROOT_DIR/data/user/common/audit-log-import/skip" 1>&3 2>&3
+  ghe-ssh "$GHE_HOSTNAME" -- "sudo touch $GHE_REMOTE_ROOT_DIR/data/user/common/audit-log-import/$flag" 1>&3 2>&3
 
   if $CLUSTER; then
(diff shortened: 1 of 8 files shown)


share/github-backup-utils/ghe-restore-audit-log
```diff
@@ -22,55 +22,54 @@ setup(){
...
-remove_complete_flag(){
-  ghe_verbose "Setting instance(s) as pending for audit log import to MySQL"
-  ghe-ssh "$GHE_HOSTNAME" -- "sudo rm -rf $GHE_REMOTE_ROOT_DIR/data/user/common/audit-log-import/complete" 1>&3 2>&3
+# Check whether the remote host is running a version where the MySQL backend
+# is supported, i.e: < 2.19
+is_mysql_supported(){
+   [ "$(version $GHE_REMOTE_VERSION)" -lt "$(version 2.19.0)" ]
+}
+
+set_remote_flag(){
```
sridharavinash: minor: what do you think of naming these more specific to audit-log, since it's mostly related to audit log import settings.
```suggestion
set_remote_audit_log_import_flag(){
```

sridharavinash: didn't realize this was in it's own `audit-log` named file, so modified the above suggestion to be more explicit what we're setting.

juruen: That was a helper function that is used by the two below. I added more doc to the functions to clarify their usage.

PR #524, opened 2019-11-13
Cater for more explicit gist paths used in routes file

lildude:
When backing up a cluster, the parallel `rsyncs` used for backing up the gists can result in `File exists (17)` errors. This happens when the same Gist prefix is used in the on-disk repo path for multiple gists on multiple nodes.

A change to better handle this is coming in the next releases of GitHub Enterprise Server which slightly changes the paths sent in the routes from the GitHub Enterprise Server being backed up. This PR implements a function to cater for this in the verification that happens immediately after all repositories and gists have been backed up.

/cc @djdefi @jatoben @maclarel @github/pe-hadr @github/backup-utils @dbussink

PR #526, opened 2019-11-14
Update verify routes

jchamber904:


lildude:
Was this intentionally opened in this repo? I ask as it's a 💯 duplicate of https://github.com/github/backup-utils/pull/524

I suspect you may have wanted this against your own fork. 😄

jchamber904:
It was not intentional.  I had thought that I had only pushed to my fork.  We needed to bring these patches into backup-utils v2.18.0 in order to support our GHE v2.16.16 instance since backup-utils v2.19.0 broke us.

PR #532, opened 2019-12-16
Use sentinel file to identify type of mysql backup

jianghao0718:
For customer, we like them to be able to restore old logical backup without much effort to reset the `mysql.backup.binary` feature flag. This means following assumptions for backup-util

1. Feature flag `mysql.backup.binary` decides if a new backup taken is a binary backup or logical backup, it doesn't affect logic how we restore backup
2. Restore logic is decided by existence of sentinel file `mysql-binary-backup-sentinel` under snapshot dir. We always assume when customer restore backup using `ghe-restore`, the backup is taken by `ghe-backup`, no matter it is binary or logical backup (depending on `mysql.backup.binary`)
3. `ghe-import-mysql` still uses `mysql.backup.binary` feature flag to decide if it will restore backup as logical or binary. For example, even we restore from a logical backup, the replication configuration will restore from a binary backup from primary if `mysql.backup.binary` is set.
4. We will fail the restore if `ghe-import-mysql` is not set and we try to restore from a binary backup.
5. If `mysql.backup.binary` is not set, we still call `ghe-import-mysql` for the reason of back-compat (older version of GHES does not have new script). Also new version `ghe-import-mysql` will use logical restore script if `mysql.backup.binary` not set.

Tests
- [x] Restore from binary backup with `mysql.backup.binary=true`
- [x] Restore from logical backup with `mysql.backup.binary=false`
- [x] Restore from logical backup with `mysql.backup.binary=true`
- [x] Restore from binary backup with  `mysql.backup.binary=true` --> this suppose to fail

TODO
- [ ] Update current binary backups on test storage (AWS S3) with sentinel file after this change merges

/cc @sbryant @shlomi-noach @github/ghes-infrastructure

### DIFF
--- a/share/github-backup-utils/ghe-restore-mysql
+++ b/share/github-backup-utils/ghe-restore-mysql
@@ -43,14 +43,22 @@ is_binary_backup_feature_on(){
   ghe-ssh "$GHE_HOSTNAME" ghe-config --true "mysql.backup.binary"
 }
 
-if is_binary_backup; then
-  if ! is_binary_backup_feature_on; then
+if is_binary_backup_feature_on; then
+  # Feature "mysql.backup.binary" is on, which means new backup scripts are available
+  if is_binary_backup; then
+    IMPORT_MYSQL=ghe-import-mysql-xtrabackup
+  else
+    IMPORT_MYSQL=ghe-import-mysql-mysqldump
+  fi
+else
+  # We do not allow to restore binary backup without "mysql.backup.binary" set 
+  if is_binary_backup; then
     echo "To restore from a binary backup, you have to set ghe-config \"mysql.backup.binary\" to true" >&2  
     exit 2
+  else
+    # legacy mode
+    IMPORT_MYSQL=ghe-import-mysql
   fi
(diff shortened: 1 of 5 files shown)


share/github-backup-utils/ghe-restore-mysql
```diff
@@ -33,13 +33,32 @@ cleanup() {
...
+}
+
+if is_binary_backup; then
+  if ! is_binary_backup_feature_on; then
+    echo "To restore from a binary backup, you have to set ghe-config \"mysql.backup.binary\" to true" >&2  
+    exit 2
+  fi
+  IMPORT_MYSQL=ghe-import-mysql-xtrabackup
+else
+  IMPORT_MYSQL=ghe-import-mysql-mysqldump
```
jianghao0718: oops.. Seems this will break back-compat since legacy ghes won't have `ghe-import-mysql-mysqldump` / `ghe-import-mysql-xtrabackup`. We need to stick with `ghe-import-mysql`

jianghao0718: Fixed with extra checking for feature flag

PR #536, opened 2020-01-21
Add support for backing up repositories via gitbackups

cainejette:
Adding support for backing up repositories via gitbackups.


cainejette:
This will call the script that https://github.com/github/github/pull/132911 introduces

bin/ghe-backup
```diff
@@ -197,8 +197,18 @@ ghe-backup-audit-log || failures="$failures audit-log"
...
+# if we should use gitbackups to backup repositories
+should_use_gitbackups_for_repositories(){
+  # TODO: Confirm re-using this setting is OK vs introducing a new setting
+  ghe-ssh "$GHE_HOSTNAME" ghe-config --true "app.github.gitbackups"
+}
+
 echo "Backing up Git repositories ..."
-ghe-backup-repositories || failures="$failures repositories"
+if should_use_gitbackups_for_repositories; then
+  ghe-backup-repositories-gitbackups || failures="$failures repositories"
```
ipmsteven: I am not familiar with bach script, I am wondering if we need to make `ghe-backup-repositories-gitbackups` return different exit code in order to make the failure case work?

DangMH:
Verified the behaviour running against the dev server.

The logging was apparent and bubbled up to the console.

DangMH:
@github/ghes-infrastructure

bin/ghe-backup
```diff
@@ -197,8 +197,18 @@ ghe-backup-audit-log || failures="$failures audit-log"
 echo "Backing up hookshot logs ..."
 ghe-backup-es-hookshot || failures="$failures hookshot"
 
+# if we should use gitbackups to backup repositories
+should_use_gitbackups_for_repositories(){
```
cainejette: we need to think through this a bit more, what if the script isn't on the machine for instance (or is this protected against in some other way?)

bin/ghe-backup
cainejette: correct. if we can do that, then this will plug into the existing framework for error handling, which has repercussions for marking the backup as complete, as well as logging the steps that failed.

bin/ghe-backup
```diff
@@ -81,6 +81,25 @@ if [ "$(ls -il dest1/testfile | awk '{ print $1 }')" != "$(ls -il dest2/testfile
...
+
+# check that the appliance supports using gitbackups for repositories
+can_use_gitbackups_for_repositories(){
+  ghe-ssh "$GHE_HOSTNAME" test -e /data/github/current/bin/backup-repositories
+}
+
+# Exit early if the appliance is missing script to backup repositories using gitbackups
+if should_use_gitbackups_for_repositories; then
+  if ! can_use_gitbackups_for_repositories; then
+    echo "Error: Configuration setting 'app.github.gitbackups' is enabled but gitbackups cannot be used to back up repositories via 'ghe-backup'."
```
cainejette: maybe something like this?

```suggestion
    echo "Error: Configuration setting 'app.github.gitbackups' is enabled but this version of GHES cannot use gitbackups to back up repositories via 'ghe-backup'."
```

share/github-backup-utils/ghe-backup-repositories-gitbackups
```diff
@@ -0,0 +1,17 @@
...
+#/ ghe-backup.
+set -e
+
+# Bring in the backup configuration
+# shellcheck source=share/github-backup-utils/ghe-backup-config
+. "$( dirname "${BASH_SOURCE[0]}" )/ghe-backup-config"
+
+bm_start "$(basename $0)"
+
+echo "github-env ./bin/backup-repositories" | ghe-ssh "$GHE_HOSTNAME" -- /bin/bash
```
cainejette: We need to capture the fact that repos may have failed to backup and exit 1, so that `ghe-backup` knows to mark `repositories` step as failed in the overall process

oakeyc: Verified the logging with the ruby script updates in https://github.com/github/github/pull/132911/files

oakeyc:
Can we get some 👀  @github/backup-utils ?

PR #541, opened 2020-01-30
Fix `ghe-backup-repositories` performance for large instances

piki:
#524 introduced a performance regression in `ghe-backup-repositories`.  This PR speeds up the code introduced in #524 by roughly 3000x.  It also fixes a future bug, in which GitHub Enterprise versions like 2.20.x and 3.x.y would have received the deprecated behavior.  Finally, the PR adds performance and correctness tests to verify the changes.

**Performance:** Prior to the PR, the `parse_paths` step took 80 seconds to process 10k lines, due to `parse_paths` forking and executing `awk`, `grep`, and sometimes `dirline` for each line of the file.  At one large customer, this added over three hours to `ghe-backup` invocations, just doing text processing in the shell.  We believe other customers will be affected as well, proportional to the number of repos, wikis, and gists they are backing up.

We eliminated the redundant calls to `version` (which invokes `awk`), which brought the 10k-line time down to 20 seconds.  Replacing the `grep` call with a shell-intrinsic string match brought the time under 10 seconds.  Still not good enough.

So ultimately we rewrote `parse_paths` as an inline Ruby script, to completely eliminate fork and exec calls.  Each 10k lines now takes around 0.03 seconds, which will remove all but ~3 seconds of the 3-hour overhead at the customer who reported the bug.

### DIFF
--- a/share/github-backup-utils/ghe-backup-config
+++ b/share/github-backup-utils/ghe-backup-config
@@ -306,21 +306,23 @@ version() {
 }
 
 fix_paths_for_ghe_version() {
-  local version_GHE_REMOTE_VERSION=$(version $GHE_REMOTE_VERSION)
-  local version_2_16_23=$(version 2.16.23)
-  local version_2_17_14=$(version 2.17.14)
-  local version_2_18_8=$(version 2.18.8)
-  local vesrion_2_19_3=$(version 2.19.3)
-  
-  while read -r line; do
-    if [[ "$GHE_REMOTE_VERSION" =~ 2.16. && "$version_GHE_REMOTE_VERSION" -ge "$version_2_16_23" ]] || \
-       [[ "$GHE_REMOTE_VERSION" =~ 2.17. && "$version_GHE_REMOTE_VERSION" -ge "$version_2_17_14" ]] || \
-       [[ "$GHE_REMOTE_VERSION" =~ 2.18. && "$version_GHE_REMOTE_VERSION" -ge "$version_2_18_8" ]] || \
-       [[ "$GHE_REMOTE_VERSION" =~ 2.19. && "$version_GHE_REMOTE_VERSION" -ge "$vesrion_2_19_3" ]] && \
-       [[ "$line" =~ gist ]]; then
-      echo "$line"
-    else
-      dirname "$line"
-    fi
-  done
+  ruby -e '
+    ghe_remote_version = ARGV.first
+    version_array = ghe_remote_version.split(".").map(&:to_i)
+    fail unless version_array.first >= 2
+    is_modern = version_array[0] > 2 ||
+      (version_array[0] == 2 && version_array[1] > 19) ||
+      (version_array[0] == 2 && version_array[1] == 19 && version_array[2] >= 3) ||
+      (version_array[0] == 2 && version_array[1] == 18 && version_array[2] >= 8) ||
+      (version_array[0] == 2 && version_array[1] == 17 && version_array[2] >= 14) ||
+      (version_array[0] == 2 && version_array[1] == 16 && version_array[2] >= 23)
+
+    $stdin.each_line do |line|
+      if is_modern && line =~ /gist/
+        puts line
+      else
+        puts File.dirname(line)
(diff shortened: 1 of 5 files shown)


piki:
cc coauthors @ewgenius and @oakeyc 
cc @lildue for review as the historical owner of much of this code

share/github-backup-utils/ghe-backup-config
```diff
@@ -304,3 +304,28 @@ ghe_debug() {
 version() {
   echo "${@#v}" | awk -F. '{ printf("%d%03d%03d%03d\n", $1,$2,$3,$4); }';
 }
+
+# The list of gists returned by the source changed in 2.16.23, 2.17.14,
+# 2.18.8, and 2.19.3.  So we need to account for this difference here.
+fix_paths_for_ghe_version() {
+  ruby -e '
```
snh: This'll introduce a dependency on Ruby for the backup-host, so will require updates to https://github.com/github/backup-utils/blob/master/docs/requirements.md#backup-host-requirements.

share/github-backup-utils/ghe-backup-repositories
```diff
@@ -366,25 +366,8 @@ bm_end "$(basename $0) - Special Data Directories Sync"
...
-        echo "$line"
-      else
-        dirname "$line"
-      fi
-    done
-  }
-
   cat $tempdir/*.rsync | sort | uniq > $tempdir/source_routes
-  (cd $backup_dir/ && find * -mindepth 5 -maxdepth 6 -type d -name \*.git | parse_paths | sort | uniq) > $tempdir/destination_routes
+  (cd $backup_dir/ && find * -mindepth 5 -maxdepth 6 -type d -name \*.git | fix_paths_for_ghe_version | sort | uniq) > $tempdir/destination_routes
```
snh: A minor optimisation suggestion:

```suggestion
  (cd $backup_dir/ && find * -mindepth 5 -maxdepth 6 -type d -name \*.git | sort | uniq | fix_paths_for_ghe_version) > $tempdir/destination_routes
```

piki: Thanks for the nudge.  You're isn't faster, but this is: `s/sort | uniq/sort -u/`.  If you tell `sort` it can discard all the duplicate keys, then it has considerably less data to work with and can use both less RAM and less CPU.  We may be able to get this step _under_ the 100 seconds it took prior to #524.

piki: Eh, maybe just on Mac.  I tried it on Linux and didn't get the same result.

However, another trick that applies here: `uniq | sort | uniq`.  The data we're sorting has sequences of repeated directories any time there are forks in a network.  Running `uniq` as a pre-filter reduces the amount of data that `sort` has to sort.  We do still need the `uniq` at the end, in case the duplicates aren't (fully) clustered in the input.

snh: 👍 That looks like a great approach, I hadn't considered using `sort` twice like that.

lildude:
I’ll take a closer look later when I’m not on mobile but one big issue with this PR: it introduces a requirement on Ruby which is currently not documented. This will possibly be a major breaking unexpected and undocumented change for our customers.

test/test-ghe-backup.sh
```diff
@@ -344,3 +345,74 @@ begin_test "ghe-backup missing directories or files on source appliance"
     verify_all_backedup_data
 )
 end_test
+
+if [ "$(uname)" == "Darwin" ]; then
+    timeout() {
+        ruby -rtimeout -e 'duration = ARGV.shift.to_i; Timeout::timeout(duration) { system(*ARGV) }' "$@"
+    }
+fi
```
lildude: 🤔 this is pulling in a pretty large (time, effort and bytes of code) dependency on Ruby (which won't ship with macOS in future) just to run tests on macOS hosts which may not have `coreutils` installed. We're already assuming users have `coreutils` installed because of our Linux assumptions. How about making `coreutils` a macOS requirement too and remove this function?

PR #548, opened 2020-02-05
Solve a few issue with binary backup

jianghao0718:
In this PR, we are trying to solve a few issues when backup and restore with binary backups

1. Use pigz instead of gzip in backup-util
Backup-util uses gzip which is super slow to compress large database files. It becomes the performance bottleneck when we backup binary backup. We will use pigz instead which is multi-threaded and way faster. 

2. Backup-util tool should always pointing to sql master for binary backup restore
This is a bug in current binary backup restore. Backup-util restores mysqldump from any instance in cluster because it connects to sql master through connection. Binary backup restore has different mechanism. We need to connect to the Sql master node and copy files there instead. Customer does not necessarily restore on the MySql Master node.

/cc @github/ghes-infrastructure


djdefi:
I believe this introduces a new requirement to backup-utils, so we need to carefully consider this change. It would also need to be added to the requirements documentation here.

jianghao0718:
@djdefi Right. And the tricky thing is changing the way we compress /decompress will have back-compat issue. I will continue working on this today to make sure we have least impact possible when improving performance

djdefi:
@snh corrected me, the `pigz` happens on the GHES instance, so this does _not_ add a new dependency to the backup-utilities script after all.

jianghao0718:
@djdefi Correct. For the two code changes I had here, it is all running on GHE instances which has pigz already.

jianghao0718:
Hi @lildude and @djdefi , I added additional changes to `backup-util` for the issue we discovered yesterday, please help to review this additional change. Thanks a lot!

bin/ghe-backup
```diff
@@ -200,10 +200,15 @@ is_binary_backup(){
 
 echo "Backing up MySQL database ..."
 bm_start "ghe-export-mysql"
-echo 'set -o pipefail; ghe-export-mysql | gzip' |
+export_command="ghe-export-mysql"
+if ! is_binary_backup; then
+  # binary backup is already compressed
+  export_command+=" | pigz"
+fi
+echo "set -o pipefail; ${export_command}" |
```
lildude: I don't think we need `${var}` interpolation here.  https://github.com/github/backup-utils/blob/master/STYLEGUIDE.md#use-var-for-interpolation-only-when-required has more info on the consistency we've tried to keep.

jianghao0718:
@djdefi / @snh -- Do we always ask customer to run `ghe-restore` on their MySql master? @lildude told me he had the impression we did that. I have seen Apple run `ghe-restore` on some other appliance without Mysql service and they hit the issue 2 I mentioned in description. Although the old logical backup restore could work on the other appliances because it connects to master and run scripts via MySql client.

share/github-backup-utils/ghe-restore-mysql
```diff
@@ -43,12 +38,33 @@ is_binary_backup_feature_on(){
   ghe-ssh "$GHE_HOSTNAME" ghe-config --true "mysql.backup.binary"
 }
 
+if ghe-ssh "$GHE_HOSTNAME" test -f /etc/github/cluster ; then
+  ghe_mysql_master=$(ghe-ssh "$GHE_HOSTNAME" ghe-config "cluster.mysql-master")
+  if [ -z $ghe_mysql_master ]; then
+    ghe_mysql_master=$GHE_HOSTNAME
+  else
+    port=$(ssh_port_part "$GHE_HOSTNAME")
+    ghe_mysql_master=$ghe_mysql_master${port:+:$port}
```
lildude: I'm pretty sure we recommend/require restoring via the MySQL master in Cluster envs, but my memory could be a little rusty here.

Additionally, I don't think you need to do the port determination here. `ghe-ssh` does it: https://github.com/github/backup-utils/blob/178da66f3cbefbb16b0909cc3ed8310f47bfb0d1/share/github-backup-utils/ghe-ssh#L53-L55

jianghao0718: I did that because `$ghe_mysql_master` doesn't have ports. If a port different from default is specified, `$ghe_mysql_master` won't be able to get it

bin/ghe-backup
jianghao0718: fixed, thanks!

PR #550, opened 2020-02-10
Rc2.19.3 binary backup restore

oakeyc:
Move from: https://github.com/github/backup-utils/pull/548 so we could base the changes off of stable. Based on information from @piki 

Original comment:
In this PR, we are trying to solve a few issues when backup and restore with binary backups

1.  Use pigz instead of gzip in backup-util
Backup-util uses gzip which is super slow to compress large database files. It becomes the performance bottleneck when we backup binary backup. We will use pigz instead which is multi-threaded and way faster. (3 to 4 times faster for backing up large database)

2.  Backup-util tool should always pointing to sql master for binary backup restore
This is an issue in current binary backup restore. Backup-util restores mysqldump from any instance in cluster because it connects to sql master through connection. Binary backup restore has different mechanism. We need to connect to the Sql master node and copy files there instead. Customer does not necessarily restore on the MySql Master node. (@lildude mentioned that we tell customers to restore from the MySQL master node, but seems the old logical backup restore works with non-master node so make this change to make sure it maintains same behavior)

/cc @github/ghes-infrastructure


jianghao0718:
I think you need to bump version to 2..19.3 for `backup-utils/share/github-backup-utils/version`. @lildude  / @piki to confirm.

piki:
Please don't merge this PR.  #551 was the right way to get these commits reviewed for a release. #548 is the right vehicle to get it into `master`.  When you run `script/release`, it will create and immediately merge a PR to do the same thing this PR wants to do -- merge the commits into `stable`.

PR #558, opened 2020-02-11
Bump version: 2.19.3

jianghao0718:
Includes general improvements & bug fixes
* Fix two issues with binary backup (slow gzip compression and support for restore from instances other than SQL master) #551

/cc @github/backup-utils


onyea (commented):
Release

PR #567, opened 2020-02-13
Fix the way we connect to mysql master using ssh forwarding for binary backups

jianghao0718:
We tried to fix the issue customer uses binary backup restore by connecting to non-Sql master instance in #548 , however, the host where customer runs might not have hostname resolved like the ghes instances. We need to use ssh forwarding to connect to other ghes instances.

### DIFF
--- a/share/github-backup-utils/ghe-restore-mysql
+++ b/share/github-backup-utils/ghe-restore-mysql
@@ -42,7 +42,8 @@ ssh_config_file_opt=""
 if $CLUSTER ; then
   ghe_mysql_master=$(ghe-ssh "$GHE_HOSTNAME" ghe-config "cluster.mysql-master")
   if [ -z $ghe_mysql_master ]; then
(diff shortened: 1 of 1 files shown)


share/github-backup-utils/ghe-restore-mysql
```diff
@@ -38,11 +38,17 @@ is_binary_backup_feature_on(){
   ghe-ssh "$GHE_HOSTNAME" ghe-config --true "mysql.backup.binary"
 }
 
-if ghe-ssh "$GHE_HOSTNAME" test -f /etc/github/cluster ; then
+ssh_config_file_opt=
+if $CLUSTER ; then
```
oakeyc: Is this already set somewhere?

jianghao0718: Yes, https://github.com/github/backup-utils/blob/473d64b4b1068f42399a80072f40c6330f4fd8ee/bin/ghe-restore#L124

PR #581, opened 2020-02-20
In legacy mode we should use ghe-import-mysql

jianghao0718:
Regression found by @lildude 

In legacy mode we should use `ghe-import-mysql` since  `ghe-import-mysql-mysqldump` doesn't exist for 2.18 / 2.19.4 or earlier

/cc @lildude @oakeyc @DanRigby @ipmsteven

### DIFF
--- a/share/github-backup-utils/ghe-restore-mysql
+++ b/share/github-backup-utils/ghe-restore-mysql
@@ -79,7 +79,7 @@ else
     exit 2
   else
     # legacy mode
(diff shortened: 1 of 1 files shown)


DanRigby:
Original comment from @lildude https://github.com/github/ghes-infrastructure/issues/278#issuecomment-589172153

jianghao0718:
Test ran against this topic branch with 2.18 and 2.19 to ensure the back-compatibility

PR #584, opened 2020-02-21
Backport mysql fix to patch release

cainejette:


jianghao0718:
The only concern I have is release script will update the version in stable. We probably need to revert back to 2.20.2 after this releases. Other than that LGTM.

PR #589, opened 2020-03-11
Compress remote_routes_list on the source instance

lowply:
There was a case where `$remote_routes_list` in `ghe-backup-storage` gets more than 140MB on a GitHub Enterprise Server instance that has tons of storage objects in the `/data/user/storage` directory. We found that out when a user enabled `GHE_DEBUG` and sent GitHub Support a huge output log file. This PR is to reduce the network traffic by `gzip` ing `$remote_routes_list` on the source instance, then `gzip -d` on the backup server.

### DIFF
--- a/share/github-backup-utils/ghe-backup-repositories
+++ b/share/github-backup-utils/ghe-backup-repositories
@@ -128,7 +128,7 @@ ghe-ssh "$GHE_HOSTNAME" -- cat $remote_routes_list | ghe_debug
 bm_end "$(basename $0) - Generating routes"
 
 bm_start "$(basename $0) - Fetching routes"
-ghe-ssh "$GHE_HOSTNAME" -- cat $remote_routes_list > $routes_list
+ghe-ssh "$GHE_HOSTNAME" -- gzip -c $remote_routes_list | gzip -d > $routes_list
 cat $routes_list | ghe_debug
 bm_end "$(basename $0) - Fetching routes"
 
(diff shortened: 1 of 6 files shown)


lowply:
The CI failed with the following error message.

```
Error: coreutils 8.31 is already installed
To upgrade to 8.32, run `brew upgrade coreutils`.
The command "brew install coreutils" failed and exited with 1 during .
```

So I believe the failure is not related to this change?

lowply:
I opened another PR (https://github.com/github/backup-utils/pull/592) and it should be reviewed / merged before this PR.

oakeyc:
Hey @lowply  I'm just going to add the work-in-progress tag on this PR until it's ready to go! Let us know when it's ready again for review :)

Clancy567:
Thank you my friend

lowply:
This is finally ready for review 👀 . @github/ghes-infrastructure Can you please have a look?

cainejette:
@lowply just making sure, you've run this and successfully backed up/restored, right?

lowply:
@cainejette Yes. I've tested backup / restore on my test GitHub Enterprise Server 2.20.7 instance. In addition to that, while running the backup / restore, I ran these commands to make sure the content of the `remote_routes_list` and the `routes_list` files are text files, not gzipped binary.

Target host:

```
while true; do sudo find /tmp -name "remote_routes_list" | xargs cat; sleep 0.1; done
```

Backup host:

```
while true; do sudo find /tmp -name "routes_list" | xargs cat; sleep 0.1; done
```

PR #592, opened 2020-03-26
Make ghe_debug handle piped stdin

lowply:
The problem this pull request is trying to solve is the same with https://github.com/github/backup-utils/pull/589, but this is much more important.

It turns out that the customer's `$remote_routes_list` was more than hundreds of MB, or possibly more than a GB. The current `ghe_debug` function receives the list as an argument. So if the input is huge, it resulted in a memory leak in bash.

This PR is to make the `ghe_debug` function able to receive data via stdin (`< /dev/stdin`) via pipe (`-p /dev/stdin`) then `while` loop to process the text data line by line.

@catto and I did a quick test and this function successfully output GB of text file transferred from the target instance.

Once this PR gets merged, I'll fix https://github.com/github/backup-utils/pull/589 to remove some conflicts.

Please note that this change won't break the current `ghe_debug` behavior. It will handle arguments as it does now.


lowply:
@lildude Hi 👋 Would you be able to review this PR? One of the failures has been resolved by providing this patch to the user.

lildude:
I'm going to defer the review to the @github/ghes-infrastructure peeps as they're the primary maintainers now and are actively working on other updates that may be related.

ryansimmen:
@lowply Looks good. Please add a test for this new behavior in [test-ghe-backup-config.sh](https://github.com/github/backup-utils/blob/master/test/test-ghe-backup-config.sh) and then I will approve. Thank you.

lowply:
@ryansimmen 👋 The test added in https://github.com/github/backup-utils/pull/592/commits/86d1ceb26bb3b56b57cb7087deccde64370d51c1. Could you have a look at it? Thank you!

lowply:
@ryansimmen 👋 Just a friendly reminder. Any chance I get a review for the test?

PR #595, opened 2020-03-30
Add GHE_BACKUP_SKIP_PAGES and GHE_RESTORE_SKIP_PAGES config settings

DanRigby:
If `GHE_BACKUP_SKIP_PAGES` is set to 'yes', `ghe-backup` will omit the backup of generated Pages artifacts.

If `GHE_RESTORE_SKIP_PAGE` is set to 'yes', `ghe-restore` will omit the restore of Pages artifacts contained in the backup and will rebuild them instead.

### DIFF
--- a/bin/ghe-backup
+++ b/bin/ghe-backup
@@ -189,7 +189,7 @@ ghe-backup-es-hookshot || failures="$failures hookshot"
 echo "Backing up Git repositories ..."
 ghe-backup-repositories || failures="$failures repositories"
 
-if [ "$GHE_RESTORE_SKIP_PAGES" = "yes" ]; then
+if [ "$GHE_BACKUP_SKIP_PAGES" = "yes" ]; then
   echo "Skipping backup of GitHub Pages artifacts ..."
 else
   echo "Backing up GitHub Pages artifacts ..."
--- a/bin/ghe-backup
+++ b/bin/ghe-backup
@@ -5,10 +5,9 @@
 #/ the MySQL database, instance settings, GitHub Pages data, etc.
 #/
 #/ OPTIONS:
-#/   -v | --verbose       Enable verbose output.
-#/   -h | --help          Show this message.
-#/        --version       Display version information.
-#/        --skip-pages    Skip backup of GitHub Pages artifacts.
+#/   -v | --verbose    Enable verbose output.
+#/   -h | --help       Show this message.
+#/        --version    Display version information.
 #/
 
 set -e
@@ -26,10 +25,6 @@ while true; do
       ;;
     -v|--verbose)
       export GHE_VERBOSE=true
-      shift
-      ;;
-    --skip-pages)
-      export GHE_SKIP_PAGES=true
       shift
       ;;
     -*)
@@ -194,11 +189,11 @@ ghe-backup-es-hookshot || failures="$failures hookshot"
 echo "Backing up Git repositories ..."
 ghe-backup-repositories || failures="$failures repositories"
 
-if [ -z "$GHE_SKIP_PAGES" ]; then
+if [ "$GHE_RESTORE_SKIP_PAGES" = "yes" ]; then
+  echo "Skipping backup of GitHub Pages artifacts ..."
+else
   echo "Backing up GitHub Pages artifacts ..."
   ghe-backup-pages || failures="$failures pages"
-else
-  echo "Skipping backup of GitHub Pages artifacts ..."
 fi
 
 echo "Backing up storage data ..."
--- a/bin/ghe-backup
+++ b/bin/ghe-backup
@@ -5,9 +5,10 @@
 #/ the MySQL database, instance settings, GitHub Pages data, etc.
 #/
 #/ OPTIONS:
-#/   -v | --verbose    Enable verbose output.
-#/   -h | --help       Show this message.
-#/        --version    Display version information.
+#/   -v | --verbose       Enable verbose output.
+#/   -h | --help          Show this message.
+#/        --version       Display version information.
+#/        --skip-pages    Skip backup of GitHub Pages artifacts.
 #/
 
 set -e
@@ -25,6 +26,10 @@ while true; do
       ;;
     -v|--verbose)
       export GHE_VERBOSE=true
+      shift
+      ;;
+    --skip-pages)
+      export GHE_SKIP_PAGES=true
       shift
       ;;
     -*)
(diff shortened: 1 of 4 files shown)


oakeyc:
Updated with the changes to restore to have the skip pages param and to gracefully exit when the backup had no pages.

bin/ghe-restore
```diff
@@ -9,21 +9,22 @@
...
-#/                     instance. The <host> may be omitted when the
-#/                     GHE_RESTORE_HOST config variable is set in backup.config.
-#/                     When a <host> argument is provided, it always overrides
-#/                     the configured restore host.
+#/   -f | --force         Don't prompt for confirmation before restoring.
+#/   -c | --config        Restore appliance settings and license in addition to
+#/                        datastores. Settings are not restored by default to
+#/                        prevent overwriting different configuration on the
+#/                        restore host.
+#/        --skip-pages    Skip backup of GitHub Pages artifacts.
```
lildude:
```suggestion
#/        --skip-pages    Skip restore of GitHub Pages artifacts.
```

DanRigby: The flag here is causing the backup script to not include the pages artifacts, so the current wording seems correct?.
We're planning on adding an equivalent flag added to `ghe-restore`.

lildude: But you've just added it to `ghe-restore` 😁

share/github-backup-utils/ghe-restore-pages
```diff
@@ -23,7 +23,17 @@ GHE_HOSTNAME="$1"
 : ${GHE_RESTORE_SNAPSHOT:=current}
 
 # Find the pages to restore
-pages_paths=$(cd $GHE_DATA_DIR/$GHE_RESTORE_SNAPSHOT/ && find pages -mindepth 5 -maxdepth 5 | cut -d / -f2-)
+
```
cainejette:
```suggestion
```

share/github-backup-utils/ghe-restore-pages
```diff
@@ -23,7 +23,17 @@ GHE_HOSTNAME="$1"
...
+check_pages=$(cd $GHE_DATA_DIR/$GHE_RESTORE_SNAPSHOT/ && find pages -mindepth 5 -maxdepth 5 2> /dev/null)
+
+# Check if there were pages in backup, if no skip restore
```
cainejette: What's the functional difference in splitting this out from the next check? They both exit early it looks like -- just a more pertinent echo message? Do we ever expect this one to trigger but not the next one?

bin/ghe-restore
```diff
@@ -9,21 +9,22 @@
...
-#/                     instance. The <host> may be omitted when the
-#/                     GHE_RESTORE_HOST config variable is set in backup.config.
-#/                     When a <host> argument is provided, it always overrides
-#/                     the configured restore host.
+#/   -f | --force         Don't prompt for confirmation before restoring.
+#/   -c | --config        Restore appliance settings and license in addition to
+#/                        datastores. Settings are not restored by default to
+#/                        prevent overwriting different configuration on the
+#/                        restore host.
+#/        --skip-pages    Skip restore of GitHub Pages artifacts.
```
cainejette: I think @itillis was hoping this could be figured out dynamically at restore time instead of requiring users to pass the flag in again. If pages aren't present in the backup (intentionally or not), we can trigger a rebuild -- are we trying to avoid that unnecessary rebuild command in the case where there actually aren't any to rebuild?

share/github-backup-utils/ghe-restore-pages
oakeyc: Yeah, so logically they are a little different in scenarios. Our change is when the backup skipped having pages altogether (there is no pages dir which caused an ugly error), but the change from before was there were pages in the backup, but it was empty. Maybe it's not that different for a customer, but I thought of them as two scenarios, hence different wording made sense

bin/ghe-restore
oakeyc: Yeah, so if there were no pages in backup (and if you choose to skip) we will add a rebuild step for pages (after that script is done), which will come in a separate PR.

bin/ghe-backup
```diff
@@ -5,9 +5,10 @@
 #/ the MySQL database, instance settings, GitHub Pages data, etc.
 #/
 #/ OPTIONS:
-#/   -v | --verbose    Enable verbose output.
-#/   -h | --help       Show this message.
-#/        --version    Display version information.
+#/   -v | --verbose       Enable verbose output.
+#/   -h | --help          Show this message.
+#/        --version       Display version information.
+#/        --skip-pages    Skip backup of GitHub Pages artifacts.
```
jianghao0718: Can we just put `GHE_SKIP_PAGES` in `backup.config`? I think that is the common way we set those variables. (it will be load by `ghe-backup-config`)

bin/ghe-restore
jianghao0718: Agree. Or a `sentinel` file if necessary

bin/ghe-restore
```diff
@@ -45,6 +46,10 @@ while true; do
       restore_settings=true
       shift
       ;;
+    --skip-pages)
+      export GHE_SKIP_PAGES=true
```
cainejette: Is this variable used outside of this script? That is, does it need to actually be exported?

DanRigby: We elected to follow the existing pattern used inside the script as we weren't initially sure if we'd have to pass the value through to a child script or set it via config (see @jianghao0718's comment above).

cainejette: the script has both local variables and exported variables. it seems like if it's contained to just this script, we should go with a local variable

cainejette: but yeah once we land on flag vs config file, that will say which method works

DanRigby: After discussion we've opted to use config settings instead of CLI flags

bin/ghe-backup
DanRigby: Done.

bin/ghe-restore
DanRigby: I don't think a sentinel file is needed in this case. We've moved to using config setting instead.

bin/ghe-backup
```diff
@@ -189,8 +189,12 @@ ghe-backup-es-hookshot || failures="$failures hookshot"
 echo "Backing up Git repositories ..."
 ghe-backup-repositories || failures="$failures repositories"
 
-echo "Backing up GitHub Pages ..."
-ghe-backup-pages || failures="$failures pages"
+if [ "$GHE_RESTORE_SKIP_PAGES" = "yes" ]; then
+  echo "Skipping backup of GitHub Pages artifacts ..."
```
cainejette: Looks like we added the word 'artifacts' here compared to before. Intentional?

```suggestion
  echo "Skipping backup of GitHub Pages ..."
```

DanRigby: Intentional, yes as there's a distinction between pages data and pages artifacts.
I'm not married to the term artifact but we're using it to distinguish between the generated byproducts and the source data that is always backed up.

cainejette: Good thinking

PR #597, opened 2020-03-31
Beta: Execute ghe-backup tasks in parallel

ryansimmen:
Execute long running backup tasks in parallel by leveraging [moreutils parallel](https://joeyh.name/code/moreutils).

This resolves https://github.com/github/ghes-infrastructure/issues/386


bin/ghe-backup
```diff
@@ -159,59 +156,54 @@ bm_init > /dev/null
...
-ghe-backup-pages || failures="$failures pages"
-
-echo "Backing up storage data ..."
-ghe-backup-storage || failures="$failures storage"
+# Execute long running backup jobs in parallel
+parallel_commands="
+ghe-backup-settings
+ghe-ssh \"$GHE_HOSTNAME\" -- 'ghe-export-authorized-keys' > authorized-keys.json
+ghe-ssh \"$GHE_HOSTNAME\" -- 'ghe-export-ssh-host-keys' > ssh-host-keys.tar
+ghe-backup-mysql
```
snh, line 164: MySQL must always be backed up first and on its own to prevent consistency issues.

This ensures that the underlying database records exist for the repositories, storage objects, pages etc that are backed up later in the process.

bin/ghe-backup
```diff
@@ -159,59 +156,54 @@ bm_init > /dev/null
...
+Backing up SSH authorized keys ...
+Backing up SSH host keys ...
+Backing up MySQL database ...
+Backing up Redis database ..
+Backing up audit log ...
+Backing up hookshot logs ...
+Backing up Git repositories ...
+Backing up GitHub Pages ...
+Backing up storage data ...
+Backing up custom Git hooks ..."
```
jianghao0718, line 184: I think now all verbose log are piped into `GHE_VERBOSE_LOG` which is a single file. Once we move this to parallel, I am worried about the verbose log will be all messed up and impossible to read. We probably should start to separate the verbose log file now.

Dockerfile
```diff
@@ -7,6 +7,8 @@ RUN apt-get -q -y update && \
     ca-certificates \
     ssh \
     git \
+    parallel \
+    && echo "will cite" | parallel --citation \
```
jianghao0718, line 11: Is this something customer also need to run?

ryansimmen: Nope, not any more with my latest changes.

bin/ghe-backup
ryansimmen: Each tasks logs will be outputted together one after another as tasks are completed. Tasks will NOT inline output with one another . Therefore verbose logs should maintain their readability. This is one of the primary reasons why I chose this utility.

bin/ghe-backup
jianghao0718: In addition to @snh 's point, with binary backup (xtrabackup) this will require us to have `ghe-backup-mysql` to run earlier than the rest.

bin/ghe-backup
ryansimmen: For good measure I went ahead and added the following phrase  to the beginning of each task.
`ghe_verbose "==== ghe-backup-task ===="`

lildude:
Two issues come to mind with this, and only because I've thought about this before when considering parallelisation in the past:

1. This makes the assumption that `parallel` in the path is GNU `parallel`. 
  This is a problem because GNU `parallel` is not the same as, and not compatible with, `parallel` as shipped with `moreutils`. These changes don't distinguish between the two and thus we could have some weird and unexpected failures. 
1. This is a breaking change with the only indication of it being a change hidden away in the `requirements.md`. 
  This new requirement is going to need to be made explicitly clear in the release notes, and should probably be documented in the README.md too, especially as it might not be easy to install both on the same system on some operating systems. I know Debian/Ubuntu plays nice, but I don't know about other OSes.

With both in mind, I think we need a guard really early in the backup and restore steps (can bung it in`share/github-backup-utils/ghe-backup-config` along with the other early checks) that checks for the GNU `parallel` and fails early if it can't be found.

lildude:
As an aside, as this is a breaking change, it should not ship until Backup Utils 2.21.0 at the earliest.

bin/ghe-backup
ryansimmen: I fixed this by using the following:
```
calling_script_name="$(caller | sed 's:.*/::')"
exec 3> >(sed "s/^/[$(date "+%Y-%m-%d %H:%M:%S")] $calling_script_name: /" >> "$GHE_VERBOSE_LOG")
```
The results look something like this:
```
[2020-04-01 08:50:38] ghe-backup-settings: * Transferring settings data ...
[2020-04-01 08:50:38] ghe-backup-settings: * Transferring license data ...
[2020-04-01 08:50:38] ghe-backup-settings: * Transferring management console password ...
[2020-04-01 08:50:38] ghe-backup-settings: * Transferring CA certificates ...
[2020-04-01 08:50:38] ghe-backup-settings: * Transferring UUID ...
[2020-04-01 08:50:38] ghe-backup-es-rsync: * Performing initial sync of ES indices ...
[2020-04-01 08:50:38] ghe-backup-es-rsync: receiving incremental file list
[2020-04-01 08:50:38] ghe-backup-es-rsync: ./
[2020-04-01 08:50:38] ghe-backup-es-rsync: 
[2020-04-01 08:50:38] ghe-backup-es-rsync: sent 169 bytes  received 8,585 bytes  17,508.00 bytes/sec
[2020-04-01 08:50:38] ghe-backup-es-rsync: total size is 121,190  speedup is 13.84
[2020-04-01 08:50:38] ghe-backup-es-rsync: * Disabling ES index flushing ...
[2020-04-01 08:50:38] ghe-backup-es-rsync: * Performing follow-up sync of ES indices ...
[2020-04-01 08:50:38] ghe-backup-es-rsync: receiving incremental file list
[2020-04-01 08:50:38] ghe-backup-es-rsync: 
[2020-04-01 08:50:38] ghe-backup-es-rsync: sent 162 bytes  received 8,578 bytes  17,480.00 bytes/sec
[2020-04-01 08:50:38] ghe-backup-es-rsync: total size is 121,190  speedup is 13.87
[2020-04-01 08:50:38] ghe-backup-es-rsync: * Enabling ES index flushing ...
[2020-04-01 08:50:38] ghe-backup-pages: 
[2020-04-01 08:50:38] ghe-backup-pages: * Starting backup for host: 10.0.1.134
[2020-04-01 08:50:38] ghe-backup-pages: 
[2020-04-01 08:50:38] ghe-backup-pages: * Transferring pages files ...
[2020-04-01 08:50:38] ghe-backup-pages: receiving incremental file list
[2020-04-01 08:50:38] ghe-backup-pages: ./
[2020-04-01 08:50:38] ghe-backup-pages: 
[2020-04-01 08:50:38] ghe-backup-pages: sent 35 bytes  received 226 bytes  522.00 bytes/sec
[2020-04-01 08:50:38] ghe-backup-pages: total size is 0  speedup is 0.00
[2020-04-01 08:50:38] ghe-backup-git-hooks: git-hooks environment tarballs not found. Skipping ...
[2020-04-01 08:50:38] ghe-backup-git-hooks: receiving incremental file list
[2020-04-01 08:50:38] ghe-backup-git-hooks: ./
[2020-04-01 08:50:38] ghe-backup-git-hooks: 
[2020-04-01 08:50:38] ghe-backup-git-hooks: sent 27 bytes  received 52 bytes  158.00 bytes/sec
[2020-04-01 08:50:38] ghe-backup-git-hooks: total size is 0  speedup is 0.00
```

share/github-backup-utils/ghe-backup-config
```diff
@@ -80,7 +80,8 @@ fi
 # otherwise, redirect it to /dev/null. Write verbose output to fd 3.
 if [ -n "$GHE_VERBOSE" ]; then
   if [ -n "$GHE_VERBOSE_LOG" ]; then
-    exec 3>> "$GHE_VERBOSE_LOG"
+    calling_script_name="$(caller | sed 's:.*/::')"
+    exec 3> >(sed "s/^/[$(date "+%Y-%m-%d %H:%M:%S")] $calling_script_name: /" >> "$GHE_VERBOSE_LOG")
```
jianghao0718, line 84: Can be even simipler by 
```
exec 3> >(sed "s/^/$calling_script_name: /" | ts >> "$GHE_VERBOSE_LOG")
``

Comment (ryansimmen) [reply share/github-backup-utils/ghe-backup-config]: Thanks, done.

File: bin/ghe-backup (hunk 1 above)
Comment (ryansimmen) [reply bin/ghe-backup]: I am now backup up mysql first.

File: share/github-backup-utils/ghe-backup-config [hunk 5]
```diff
@@ -66,6 +66,18 @@ for f in "$GHE_BACKUP_CONFIG" "$GHE_BACKUP_ROOT/backup.config" \
   fi
 done
 
+# Check that moreutils ts has been installed
+if ! echo "moreutils" | ts &>/dev/null; then
+  echo "Error: moreutils ts not found. Please install https://joeyh.name/code/moreutils" 1>&2
```
lildude, line 71:
```suggestion
  echo "Error: moreutils not found. Please install https://joeyh.name/code/moreutils" 1>&2
```

moreutils supplies `ts` so I don't think we don't need to call out `ts` specifically.

bin/ghe-backup
```diff
@@ -41,8 +41,16 @@ done
 # shellcheck source=share/github-backup-utils/ghe-backup-config
 . "$( dirname "${BASH_SOURCE[0]}" )/../share/github-backup-utils/ghe-backup-config"
 
+# Check that the parallel_command is pointing to moreutils parallel
+if ! $parallel_command -h | grep "parallel \[OPTIONS\] command -- arguments" &>/dev/null; then
```
lildude, line 45:
```suggestion
if ! $parallel_command -h | grep -q "parallel \[OPTIONS\] command -- arguments"; then
```

No need for redirection with `-q`.

I also suggest putting this into `share/github-backup-utils/ghe-backup-config` and combine it with where you're defining `$parallel_command` as you can't really have one without the other.

As it stands, this check would also need to be added to `ghe-restore` if you decide to implement parallel restores. Moving it now will make the `ghe-restore` get the functionality and check for free.

bin/ghe-backup
```diff
@@ -41,8 +41,16 @@ done
...
+  echo "Error: moreutils parallel not found. Please install https://joeyh.name/code/moreutils" 1>&2
+  exit 1
+fi
+
 # Used to record failed backup steps
 failures=
+failures_path="failures.txt"
```
lildude, line 52: This is making an assumption that this dir is writeable by the backup user as we've not yet changed to the snapshot dir, which won't be the case on those systems with clear separation of responsibility - ie the backup user may not necessarily be the user who installs and updates backup-utils. 

As this is only a temporary file, it's better to use a tmp location using `mktemp` and clean up the file at the end.

bin/ghe-backup
ryansimmen: I implemented `grep -q` as you suggested and moved the moreutils paralllel check into `ghe-backup-config` so it may be referenced later from `ghe-restore`.

bin/ghe-backup
ryansimmen: Fixed

share/github-backup-utils/ghe-backup-config
ryansimmen: Done

bin/ghe-backup
```diff
@@ -41,8 +41,13 @@ done
 # shellcheck source=share/github-backup-utils/ghe-backup-config
 . "$( dirname "${BASH_SOURCE[0]}" )/../share/github-backup-utils/ghe-backup-config"
 
+# Check to make sure moreutils parallel is installed
+ghe_parallel_check
```
lildude, line 45: Ooops, I realise now I wasn't clear about what I had in mind in my comment earlier today. This approach would still need to add this check the `ghe-restore`.

What I had in mind was  is if you make this function, or even just its content, exec directly within `ghe-backup-config` you kill two birds with one stone.

ryansimmen:
> 1. This makes the assumption that `parallel` in the path is GNU `parallel`.
>    This is a problem because GNU `parallel` is not the same as, and not compatible with, `parallel` as shipped with `moreutils`. These changes don't distinguish between the two and thus we could have some weird and unexpected failures.
> 
> With both in mind, I think we need a guard really early in the backup and restore steps (can bung it in`share/github-backup-utils/ghe-backup-config` along with the other early checks) that checks for the GNU `parallel` and fails early if it can't be found.

This PR is now using moreutils parallel instead of GNU parallel and if both are installed the code will check the path for the presence of parallel and choose the appropriate executable.

bin/ghe-backup
```diff
@@ -41,8 +41,13 @@ done
...
+
 # Used to record failed backup steps
 failures=
+tempdir="$(mktemp -d -t backup-utils-backup-XXXXXX)"
```
lildude, line 49: I don't think you need a whole directory as everything is going to a single file. Why not go straight for the temp file like this?

```suggestion
failures_file="$(mktemp backup-utils-backup-failures-XXXXXX)"
```

You can then remove the next line and update the var in `cleanup()`.

share/github-backup-utils/ghe-backup-config
```diff
@@ -66,6 +66,26 @@ for f in "$GHE_BACKUP_CONFIG" "$GHE_BACKUP_ROOT/backup.config" \
...
+    parallel_command="/usr/bin/parallel.moreutils"
+  fi
+
+  # Check that the parallel_command is pointing to moreutils parallel
+  if ! $parallel_command -h | grep -q "parallel \[OPTIONS\] command -- arguments"; then
+    echo "Error: moreutils not found. Please install https://joeyh.name/code/moreutils" 1>&2
+    exit 1
+  fi
+}
+
```
lildude, line 88: I wasn't very clear here earlier either, sorry. What I had in mind was combining these two together, ala:

```suggestion
# Check that moreutils has been installed
ghe_parallel_check() {
  if ! echo "moreutils" | ts &>/dev/null; then
    echo "Error: moreutils not found. Please install https://joeyh.name/code/moreutils" 1>&2
    exit 1
  fi
  
  # Some machines may have both moreutils parallel and GNU parallel installed.
  parallel_command="parallel"
  if [ -x "/usr/bin/parallel.moreutils" ]; then
    parallel_command="/usr/bin/parallel.moreutils"
  fi

  # Check that the parallel_command is pointing to moreutils parallel
  if ! $parallel_command -h | grep -q "parallel \[OPTIONS\] command -- arguments"; then
    echo "Error: moreutils not found. Please install https://joeyh.name/code/moreutils" 1>&2
    exit 1
  fi
}
```

This doesn't have to be wrapped in a function and can be exec'd directly.

ryansimmen: I tried calling these checks directly and not wrapping them in a function originally, but several of the inner scripts were failing to find `parallel` in it's path and was prompting for the installation of `moreutils`. This was causing `ghe-backup-pages`, etc. to fail. 

Plus there's no reason for every `ghe-backup-*` script that's invoked from within `ghe-backup` to execute these checks. The same especially goes for all of the `ghe-restore` scripts since restore is not using parallel at all. Wrapping these checks in a function that can be called from `ghe-restore` once we decide to go that route seems to me like the best compromise. Would you agree?

bin/ghe-backup
ryansimmen: I replied to this comment below

bin/ghe-backup
ryansimmen: Done

share/github-backup-utils/ghe-backup-config
```diff
@@ -66,6 +66,39 @@ for f in "$GHE_BACKUP_CONFIG" "$GHE_BACKUP_ROOT/backup.config" \
...
+  fi
+
+  # Check that the GHE_PARALLEL_COMMAND is pointing to moreutils parallel
+  if ! $GHE_PARALLEL_COMMAND -h | grep -q "parallel \[OPTIONS\] command -- arguments"; then
+    echo "Error: moreutils not found. Please install https://joeyh.name/code/moreutils" 1>&2
+    exit 1
+  fi
+
+  GHE_PARALLEL_COMMAND_OPTIONS=""
+  if [ "$GHE_PARALLEL_ENABLED" != "yes" ]; then
```
cainejette, line 89: This was a good idea to put it behind configuration. I assume there's a commit that just hasn't been pushed up that includes this and some comments in the `backup.config-example` file ?

share/github-backup-utils/ghe-backup-config
lildude: Ah yeah. Good point.

share/github-backup-utils/ghe-backup-config
ryansimmen: I think @itillis doesn't want us to advertise this flag in our documentation just yet.

test/test-ghe-backup-config.sh
```diff
@@ -127,16 +127,3 @@ begin_test "ghe-backup-config ghe_parse_remote_version v2.x series"
...
-
-  export GHE_VERBOSE=1
-  export GHE_VERBOSE_LOG="$TRASHDIR/verbose.log"
-  . "share/github-backup-utils/ghe-backup-config"
-  ghe_verbose "Hello world"
-  [ "$(wc -l <"$GHE_VERBOSE_LOG")" -gt 0 ]
-  unset GHE_VERBOSE
-  unset GHE_VERBOSE_LOG
-)
-end_test
```
ryansimmen, line 142: Removing this test because of buffering issues surrounding fd 3 redirection. The log will eventually get written to, but not in time for this test to successfully pass.

lildude: This delay can't be very long. Wouldn't a `sleep 1` or a `while` loop be a better option than risking the chance of breaking verbose logging without realising it and only discovering when a critical issue needs debugging?

I'll leave that judgment to you and not block this PR for this.

ryansimmen: I was able to fix this bug by leveraging `awk`'s `fflush()`
```
exec 3> >(awk -v c=$calling_script_name '{ print strftime("%b %d %H:%M:%S"), c":", $0; fflush(); }' >> "$GHE_VERBOSE_LOG")
```

I've added the test back in and it's passing.

share/github-backup-utils/ghe-backup-config
ryansimmen: I've added this flag to our documentation and marked it as a beta feature.

README.md
```diff
@@ -3,6 +3,8 @@
 This repository includes backup and recovery utilities for
 [GitHub Enterprise Server][1].
 
+**UPDATE**: `backup-utils` now requires [awk](https://www.gnu.org/software/gawk) and [moreutils](https://joeyh.name/code/moreutils).
```
lildude, line 6: There are several variants of awk so we should be explicit here.

```suggestion
**UPDATE**: `backup-utils` now requires [GNU awk](https://www.gnu.org/software/gawk) and [moreutils](https://joeyh.name/code/moreutils).
```

docs/requirements.md
```diff
@@ -6,7 +6,7 @@ storage and must have network connectivity with the GitHub Enterprise Server app
 ## Backup host requirements
 
 Backup host software requirements are modest: Linux or other modern Unix operating
-system with [bash][1], [git][2], [OpenSSH][3] 5.6 or newer, and [rsync][4] v2.6.4 or newer.
+system with [awk][10], [bash][1], [git][2], [moreutils][9], [OpenSSH][3] 5.6 or newer, and [rsync][4] v2.6.4 or newer.
```
lildude, line 9:
```suggestion
system with [GNU awk][10], [bash][1], [git][2], [moreutils][9], [OpenSSH][3] 5.6 or newer, and [rsync][4] v2.6.4 or newer.
```

ryansimmen: Done

README.md
ryansimmen: Done

ryansimmen:
Here is a thread describing why gawk is needed over mawk https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=593504

backup.config-example
```diff
@@ -61,3 +61,21 @@ GHE_NUM_SNAPSHOTS=10
...
+#
+# WARNING: this feature is in beta.
+#GHE_PARALLEL_MAX_JOBS=2
+
+# When jobs are running in parallel wait as needed to avoid starting new jobs
+# when the system's load average is not below the specified limit. Defaults to
+# unrestricted.
+#
+# WARNING: this feature is in beta.
+GHE_PARALLEL_MAX_LOAD=50
```
lildude, line 81: Shouldn't this be commented out by default?

README.md
```diff
@@ -3,6 +3,8 @@
 This repository includes backup and recovery utilities for
 [GitHub Enterprise Server][1].
 
+**UPDATE**: `backup-utils` now requires [GNU awk](https://www.gnu.org/software/gawk) and [moreutils](https://joeyh.name/code/moreutils).
```
lildude, line 6: As this PR can now run fine without `parallel`, we should probably change the wording here to reflect this is only needed for the beta functionality.

share/github-backup-utils/ghe-backup-config
```diff
@@ -66,6 +66,32 @@ for f in "$GHE_BACKUP_CONFIG" "$GHE_BACKUP_ROOT/backup.config" \
...
+  fi
+
+  # Check that the GHE_PARALLEL_COMMAND is pointing to moreutils parallel
+  if ! $GHE_PARALLEL_COMMAND -h | grep -q "parallel \[OPTIONS\] command -- arguments"; then
+    echo "Error: moreutils not found. Please install https://joeyh.name/code/moreutils" 1>&2
+    exit 1
+  fi
+
+  if [ -n "$GHE_PARALLEL_MAX_JOBS" ]; then
+    GHE_PARALLEL_COMMAND_OPTIONS="-j $GHE_PARALLEL_MAX_JOBS"
```
cainejette, line 87: Not sure we need this space, but the option just below has it -- I'm looking to see where they're used to check but wanted to note it in the mean time...

```suggestion
    GHE_PARALLEL_COMMAND_OPTIONS=" -j $GHE_PARALLEL_MAX_JOBS"
```

backup.config-example
ryansimmen: Nice catch

README.md
ryansimmen: Fixed

share/github-backup-utils/ghe-backup-config
ryansimmen: Yep this was intentional, thanks.

README.md
```diff
@@ -3,6 +3,8 @@
 This repository includes backup and recovery utilities for
 [GitHub Enterprise Server][1].
 
+**UPDATE**: The new parallel backup and restore beta feature, will require [GNU awk](https://www.gnu.org/software/gawk) and [moreutils](https://joeyh.name/code/moreutils) to be installed.
```
cainejette, line 6:
```suggestion
**UPDATE**: The new parallel backup and restore beta feature will require [GNU awk](https://www.gnu.org/software/gawk) and [moreutils](https://joeyh.name/code/moreutils) to be installed.
```

backup.config-example
```diff
@@ -61,3 +61,21 @@ GHE_NUM_SNAPSHOTS=10
...
+#
+# WARNING: this feature is in beta.
+#GHE_PARALLEL_MAX_JOBS=2
+
+# When jobs are running in parallel wait as needed to avoid starting new jobs
+# when the system's load average is not below the specified limit. Defaults to
+# unrestricted.
+#
+# WARNING: this feature is in beta.
+#GHE_PARALLEL_MAX_LOAD=50
```
cainejette, line 81: The units for this are supposed to be percent, is that right? Perhaps we can make that clearer?

docs/requirements.md
```diff
@@ -8,6 +8,8 @@ storage and must have network connectivity with the GitHub Enterprise Server app
 Backup host software requirements are modest: Linux or other modern Unix operating
 system with [bash][1], [git][2], [OpenSSH][3] 5.6 or newer, and [rsync][4] v2.6.4 or newer.
 
+The new parallel backup and restore beta feature, will require [GNU awk][10] and [moreutils][9] to be installed.
```
cainejette, line 11:
```suggestion
The new parallel backup and restore beta feature will require [GNU awk][10] and [moreutils][9] to be installed.
```

ryansimmen: Fixed

backup.config-example
ryansimmen: Fixed

README.md
ryansimmen: Fixed

randyr505:
Glad to see my POC has come to fruition! I can't wait to test this out in our environment! @ryansimmen is there a limit for the parallel rsyncs? In my POC I ran 7000 at a time, lol. Thanks for your great work on this!

ryansimmen:
> Glad to see my POC has come to fruition! I can't wait to test this out in our environment! @ryansimmen is there a limit for the parallel rsyncs? In my POC I ran 7000 at a time, lol. Thanks for your great work on this!

@randyr505 yes, you may simply set `GHE_PARALLEL_RSYNC_MAX_JOBS`. For reference please see https://github.com/github/backup-utils/blob/master/backup.config-example#L72

PR #598, opened 2020-04-01
add configuration + skip pages to usage

oakeyc:


docs/usage.md
```diff
@@ -10,7 +10,13 @@ After the initial backup, use the following commands:
 
 These commands are run on the host you [installed][1] Backup Utilities on.
 
-### Example backup and restore usage
+## Understanding Configuration
+
+Users can supply their own configuration file or use the example configuration file as a template where they can set up their environment for backing up and restoring.
+
+Among other levers, here is where you can specify if you want to skip backing up and/or reestoring pages artifacts with `GHE_BACKUP_SKIP_PAGES` and `GHE_RESTORE_SKIP_PAGES` respectively.
```
DanRigby, line 17:
```suggestion
An example configuration file with documentation on possible settings can found in [backup.config-example](../backup.config-example).
```

cainejette, line 17:
```suggestion
Among other levers, here is where you can specify if you want to skip backing up and/or restoring pages artifacts with `GHE_BACKUP_SKIP_PAGES` and `GHE_RESTORE_SKIP_PAGES`, respectively.
```

docs/usage.md
```diff
@@ -10,7 +10,13 @@ After the initial backup, use the following commands:
 
 These commands are run on the host you [installed][1] Backup Utilities on.
 
-### Example backup and restore usage
+## Understanding Configuration
```
cainejette, line 13:
```suggestion
## Configuring backup and restore behavior
```

cainejette:
We should probably explicitly mention the point @DanRigby brought up about how Pages content will be rebuilt even if the artifacts aren't restored. Or not, since that is called out in the config file itself, I suppose. What do you think?

oakeyc:
> We should probably explicitly mention the point @DanRigby brought up about how Pages content will be rebuilt even if the artifacts aren't restored. Or not, since that is called out in the config file itself, I suppose. What do you think?

I guess it depends on whether we merge this before or after that work is finished. As of the current state of the world, pages do **not** get rebuilt when they are skipped

docs/usage.md
DanRigby: This wording got replaced with a link to the example config instead since it documents the existing settings.

docs/usage.md
```diff
@@ -10,7 +10,13 @@ After the initial backup, use the following commands:
...
+
+Users can supply their own configuration file or use the example configuration file as a template where they can set up their environment for backing up and restoring.
```
cainejette, line 15: Since the main heading and the next sentence are both in second person, should we be a little less formal here?

cainejette:
> pages do *not* get rebuilt when they are skipped

Good point - we probably shouldn't have said that in the first merge.

cainejette:
> we probably shouldn't have said that in the first merge.

Thinking more about this, we should probably take out it of the config file until it's true?

oakeyc:
> > we probably shouldn't have said that in the first merge.
> 
> Thinking more about this, we should probably take out it of the config file until it's true?

I do agree that it would be cleaner to remove it until it's there. I think however, it should be coming shortly, so I'm not sure how big of a deal it would be to just leave it for the mean time. Esp, because I think visibility would be low until then anyways

PR #602, opened 2020-04-06
[WIP] add rebuild script for pages

oakeyc:


bin/ghe-restore
```diff
@@ -269,13 +269,15 @@ if $CLUSTER || [ "$(version $GHE_REMOTE_VERSION)" -ge "$(version 2.13.0)" ]; the
   ghe-restore-repositories-gist "$GHE_HOSTNAME"
 
   if [ "$GHE_RESTORE_SKIP_PAGES" = "yes" ]; then
-    echo "Skipping restore of GitHub Pages artifacts ..."
+    echo "Skipping restore of GitHub Pages artifacts. Rebuilding artifacts ..."
```
cainejette, line 272: To reinforce the expectation that this won't happen synchronously, perhaps we should include some wording like that.

"Enqueuing job to rebuild artifacts asynchronously..." or something along those lines?

PR #603, opened 2020-04-10
Run repositories restore in parallel

ewgenius:
Execute rsync in parallel (1 process per server) using [moreutils parallel](https://joeyh.name/code/moreutils)


share/github-backup-utils/ghe-restore-repositories
```diff
@@ -140,20 +143,37 @@ fi
...
-  ghe_verbose "* Transferring repository networks to $server ..."
-  ghe-rsync -avrR --delete \
-    -e "ssh -q $opts -p $port $ssh_config_file_opt -l $user" \
-    --rsync-path="sudo -u git rsync" \
-    --files-from=$file_list \
-    "$GHE_DATA_DIR/$GHE_RESTORE_SNAPSHOT/repositories/./" \
-    "$server:$GHE_REMOTE_DATA_USER_DIR/repositories/" 1>&3
+
+  rsync_commands+=("
+    if [ -n \"$GHE_VERBOSE\" ]; then
```
omgitsads, line 155: Do we want to do this some other way? Since `ghe_verbose` isn't accessible while running with `parallel` we currently have to do it this way, but it seems a little clunky.

test/test-ghe-restore-parallel.sh
```diff
@@ -4,6 +4,12 @@ set -e
 
 export GHE_PARALLEL_ENABLED=yes
 
+# use temp dir to fix rsync file issues in parallel execution:
+# we are imitating remote server by local files, and running rsync in parallel may cause
+# race conditions when two processes writing to same folder
+tempdir=$(mktemp -d -t backup-utils-restore-temp-XXXXXX)
```
ryansimmen, line 10: Use a different variable name here for good measure. tempdir is common across the code and we do a lot of sourcing unfortunately.

share/github-backup-utils/ghe-restore-repositories
```diff
@@ -140,20 +143,37 @@ fi
...
+      echo \"* Transferring repository networks to $server ($file_list) ...\" 1>&3
+      cat $file_list
```
ryansimmen, line 157: Why are we catting this? You can use ghe_verbose to send something to the verbose log.

share/github-backup-utils/ghe-restore-repositories
ryansimmen: I'm pretty certain that outputting to fd3 does work in parallel. You can safely remove this if statement.

omgitsads: Outputting to fd3 does work, however `ghe_verbose` guards on `$GHE_VERBOSE` being set prior to outputting: https://github.com/github/backup-utils/blob/9449255689b7b05a6af685891ef3fa206d92c913/share/github-backup-utils/ghe-backup-config#L325-L329

The intention was to mimic `ghe_verbose`, but I guess we can remove the if statement.

`ghe_verbose` doesn't work because we're not running these commands within the environment of this script with `parallel`, we'd need to `source ghe-backup-config` in the generated `rsync_commands` to provide access to it.

share/github-backup-utils/ghe-restore-repositories
omgitsads: 👍 This shouldn't be here at all. As I explained previously, `ghe_verbose` isn't available in the `parallel` environment

share/github-backup-utils/ghe-restore-repositories
ryansimmen: Oh I see your intention. Thanks for explaining. Let's leave it as is.

test/test-ghe-restore-parallel.sh
omgitsads: Changed this to `parallel_rsync_tempdir`.

PR #604, opened 2020-04-10
Back out Skip Pages modifications

DanRigby:
- We're not moving forward with these changes for now, so remove them from master.


cainejette:
FYI we have `github/backup-utils` now populated also :ok_hand:

rsmedes:
This feature would be nice to decrease backup/restore for larger instances. Voting that we reconsider implementing this feature.

yvettep321 (commented):
thank you

PR #609, opened 2020-05-11
add long term backup support with ghe-preserve

cvega:
We have a couple of teams on services that are interested in preserving backups long term to meet specific compliance needs. To meet this need we created a `ghe-preserve`.

Looking to get feedback to see if something like this would be useful for inclusion. 

**What it does:**
- create a date and hour range matching strategy
  - hourly snapshots for 48 hours
  - 14 days
  - 8 weeks (sunday)
  - 24 months
  - 5 years
- files matching the generated times/dates are copied to a 'preserve' folder assuming they do not already exist
- files in the preserve folder not matching hourly and dates are deleted
  - special care taken for edge cases (e.g. 48 hours can span 3 calendar days)

**Other**
- It never modifies an existing snapshot
- shellcheck mostly happy
- I did not update ghe-backup-config to include the preserve path, probably should
- I'm using arrays in a few places despite the style guide. I'd be interested in feedback and learning more in terms of avoiding this
- I didn't make it configurable because date time work coupled with deleting directories is honestly pretty scary and having a single state is easier.
- Not knowing a ton about how these snapshots work, is moving this data even viable for a restore assuming a user had to restore a snapshot from for example 18 months ago.

### DIFF
--- a/bin/ghe-preserve
+++ b/bin/ghe-preserve
@@ -1,0 +1,150 @@
+#!/usr/bin/env bash
+#/ Usage: ghe-preserve [-hv] [--version]
+#/
+#/ Preserve GitHub snapshots utilizing a general practice approach to
+#/ maintaining snapshots long term (5 years). 
+#/
+#/          +----------+----------------+-------------+
+#/          | INTERVAL | PRESERVE       | STRATEGY    |
+#/          +----------+----------------+-------------+
+#/          | hourly   | last 48 hours  | YYYYMMDDTHH |
+#/          +----------+----------------+-------------+
+#/          | daily    | last 14 days   | YYYYMMDD    |
+#/          +----------+----------------+-------------+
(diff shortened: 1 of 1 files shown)


tuanggo:
Cập nhật lên phiên bản mới

gamefiend:
Hi @cvega! Apologies for the delay in response. 

Before we go too deep, I want to check: Is this something you still need, or have your needs shifted?

cvega:
Hello @gamefiend :wave:

We've provided this to a github customer who's been running it for awhile. I'm curious if this is something the team would be interested in and what if anything it would take to incorporate.

gamefiend:
If we still want to work on this, can you submit the change to the private repo? We can absolutely look at adding this feature and can open up a discussion for it. Will close this PR in the meantime.

PR #615, opened 2020-06-09
Remove loop of testing against older versions

cainejette:
There used to be different code paths executed per GHES version, so we'd run the tests against each simulated version. We no longer have different code paths, so we're running the tests twice for no reason.


script/cibuild
```diff
@@ -6,7 +6,7 @@ set -e
 # place with this version at the beginning of each test and many commands have
 # conditional logic based on the remote version. Running the suite against
 # different major versions ensures we're covering these conditional paths.
-REMOTE_VERSIONS="2.18.0 2.20.0"
+REMOTE_VERSIONS="2.18.0 2.19.0 2.20.0"
```
ipmsteven, line 9: do we need to add 2.21.0 as well since it is releasing right now

cainejette: No. I am going to wait for the release to go in and then update my addition to be 2.20.

ipmsteven: >different major versions

maybe we could call it `different release versions`

ipmsteven: ok, may I know where is the logic we have version validation?

script/cibuild
```diff
@@ -6,7 +6,7 @@ set -e
 # place with this version at the beginning of each test and many commands have
 # conditional logic based on the remote version. Running the suite against
 # different major versions ensures we're covering these conditional paths.
-REMOTE_VERSIONS="2.19.0 2.21.0"
+REMOTE_VERSIONS="2.19.0 2.20.0 2.21.0"
```
ipmsteven, line 9: hmm, it seems we need also add 2.19 here since that is our current supported version of  ghes

https://github.com/github/backup-utils/blob/5e551d24f9fc13c40371e623ac3ac90b74ae57b5/test/bin/ghe-negotiate-version#L6

cainejette: What do you mean? 2.19 is the first one listed

lildude: This is mostly useless these days as I don't think there are any more version-specific code paths so is really just acting as a version verifier that unnecessarily runs the entire test suite twice. Adding a third version just adds another unnecessary delay.

ipmsteven: hmm, I feel there are three version constraints here:
1. The absolute versions of GHES we supported: which are 2.18, 2.19, 2.20, and 2.21
2. The absolute versions of backup-util we supported: do we have it now?
3. The relative versions constraint between `backup-util` and `GHES`, which are:
Backup Utilities `major`.`minor(or feature)`.`patch` can be used to backup and restore for GHES with version from `major`.`minor-2`.`patch` to `major`.`minor`.`patch`

My question is: do we need to have version validation for ALL version constraints mentioned above?

cainejette: Ah interesting, that is a good point. Perhaps we should just drop this altogether and only run tests once...

ipmsteven: hmm, I feel there are three version constraints here:
1. The absolute versions of GHES we support, right now it is: 2.18, 2.19, 2.20, and 2.21
2. The absolute versions of backup-util we support: do we have it now?
3. The [relative versions constraint](https://github.com/github/backup-utils/blob/master/docs/requirements.md#github-enterprise-server-version-requirements) between `backup-util` and `GHES`, which are:
Backup Utilities `major`.`minor(or feature)`.`patch` can be used to backup and restore for GHES with version from `major`.`minor-2`.`patch` to `major`.`minor`.`patch`

My question is: do we need to have version validation for ALL version constraints mentioned above in backup-util layer?

script/cibuild
```diff
@@ -2,28 +2,19 @@
...
-do
-  echo "==> Running testsuite with GHE_TEST_REMOTE_VERSION=$version"
-  export GHE_TEST_REMOTE_VERSION="$version"
-  if ! find test -name "test-*.sh" -print0 | xargs -0 -P 4 -n 1 /bin/bash; then
-    res=false
-  fi
-  echo
-done
+version=2.21.0
+echo "==> Running testsuite with GHE_TEST_REMOTE_VERSION=$version"
```
ipmsteven, line 10:
```suggestion
echo "==> Running test suite with GHE_TEST_REMOTE_VERSION=$version"
```

ipmsteven:
I am wondering why we are removing the test for old version.
if we removes it, then this existing test seems not  make sense anymore?
```
begin_test "ghe-host-check blocks restore to old release"
(
  set -e
  
  mkdir -p "$GHE_DATA_DIR/current/"
  echo "$GHE_TEST_REMOTE_VERSION" > "$GHE_DATA_DIR/current/version"

  # shellcheck disable=SC2046 # Word splitting is required to populate the variables
  read -r bu_version_major bu_version_minor bu_version_patch <<<$(ghe_parse_version $GHE_TEST_REMOTE_VERSION)
  ! GHE_TEST_REMOTE_VERSION=$bu_version_major.$((bu_version_minor-1)).$bu_version_patch ghe-restore -v
)
```

lildude:
> if we removes it, then this existing test seems not make sense anymore?

It does make sense. That test is testing to ensure we prevent restoring a backup of GHES 2.21.x to GHES 2.20.x.

script/cibuild
```diff
@@ -2,28 +2,19 @@
...
-  fi
-  echo
-done
+version=2.21.0
+echo "==> Running test suite with GHE_TEST_REMOTE_VERSION=$version"
+export GHE_TEST_REMOTE_VERSION="$version"
+if ! find test -name "test-*.sh" -print0 | xargs -0 -P 4 -n 1 /bin/bash; then
+  res=false
+fi
+echo
```
lildude, line 15: This is introducing another place we'll need to keep updating the version with each release. It's also unnecessary as the version will always be overwriting the the same default version defined at https://github.com/github/backup-utils/blob/b774394b879d55b80a3f8d00a02454fe0bb36487/test/testlib.sh#L43-L46

We can 🔥 this entire section.

```suggestion
```

cainejette: Oh! Ok, great!

script/cibuild
```diff
@@ -2,30 +2,9 @@
...
-    res=false
-  fi
-  echo
-done
-
-# If any of the version tests failed, exit non-zero
-$res
-
 # Bail out when --no-package given
 [ "$1" = "--no-package" ] && exit 0
```
cainejette, line 9: It looks like this `--no-package` flag was there to allow running the tests against different versions? Since those are now removed, perhaps this flag can be removed as well?

cainejette: Which would mean this can be removed, as well? https://github.com/github/backup-utils/blob/add-middle-version/Makefile#L3-L4

Is that right, @lildude ? I am still piecing together how this was used, so appreciate the extra set of eyes

cainejette: Looks like that's called by CI: https://github.com/github/backup-utils/blob/add-middle-version/.travis.yml#L14

script/cibuild
cainejette: I initially committed this suggestion, but then came to the conclusion following what other code might be deleted in this thread https://github.com/github/backup-utils/pull/615#discussion_r444481495 that we need to run the tests here at least once, so I removed the bits that were unnecessary but left the test execution.

Let me know what you think.

script/cibuild
cainejette: So I think we want to run the tests here, and that means leaving the `--no-package` flag

script/cibuild
lildude: 🤦 oops. Yes, we still need that line.

script/cibuild
lildude: That said, looking at the old Travis config, it looks like that flag is purely to prevent attempting to build the tarball and Debian pkg when building with `debuild` as it runs all tests before attempting to build the deb file and we don't want to include these files in the tarball or deb.

So the only time this section would be run is if manually without the flag. I'd say, keep this for now.

script/cibuild
cainejette: This has been removed in the latest iteration of the PR, so I think we're good here.

script/cibuild
cainejette: Cool, thanks for clarifying that for me!

script/cibuild
cainejette: Makes sense to me, thanks!

script/cibuild
cainejette: I don't think we _need_ to but it would certainly give us more confidence if we did!

PR #618, opened 2020-06-15
Setup GH action

ewgenius:
- switched from Travis CI to actions
- added [super-linter action](https://github.com/github/super-linter)
> note: linter for now configured to run only on new/changed files - it detects quite big amount of issues at the moment, so it is better to fix them gradually


.github/workflows/lint.yml
```diff
@@ -0,0 +1,19 @@
...
+    name: Lint Code Base
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Code
+        uses: actions/checkout@v2
+      - name: Lint Code Base
+        uses: docker://github/super-linter:v2.1.1
+        env:
+          VALIDATE_ALL_CODEBASE: false
+          VALIDATE_ANSIBLE: false
```
oakeyc, line 19: new line ?

.github/workflows/main.yml
```diff
@@ -0,0 +1,39 @@
...
+      
+    steps:
+    - name: Install Dependencies (Linux)
+      run: |
+        sudo apt-get update -y
+        sudo apt-get install -y devscripts debhelper moreutils fakeroot jq pigz
+        wget "https://storage.googleapis.com/shellcheck/shellcheck-v0.7.0.linux.x86_64.tar.xz"
+        tar --xz -xvf "shellcheck-v0.7.0.linux.x86_64.tar.xz"
+        sudo cp shellcheck-v0.7.0/shellcheck /usr/bin/shellcheck
+      if: matrix.os != 'macos-latest'
```
ipmsteven, line 22: how about 
```
if: matrix.os == 'ubuntu-latest'
```
which is consistent with the `if: matrix.os == 'macos-latest'` below

.github/workflows/main.yml
```diff
@@ -0,0 +1,39 @@
...
+    - name: Get Sources
+      uses: actions/checkout@v2
+    - name: Test
+      run: |
+        export PATH="$PATH:/snap/bin"
+        make test
+      shell: bash
+    - name: Build (Linux)
+      run: DEB_BUILD_OPTIONS=nocheck debuild -us -uc
+      if: matrix.os != 'macos-latest'
```
ipmsteven, line 38: same here  https://github.com/github/backup-utils/pull/618/files#r443738293

.github/workflows/lint.yml
ipmsteven, line 19: nit: add a new line at the end of this file

.github/workflows/main.yml
```diff
@@ -0,0 +1,39 @@
...
+      uses: actions/checkout@v2
+    - name: Test
+      run: |
+        export PATH="$PATH:/snap/bin"
+        make test
+      shell: bash
+    - name: Build (Linux)
+      run: DEB_BUILD_OPTIONS=nocheck debuild -us -uc
+      if: matrix.os != 'macos-latest'     
+
```
ipmsteven, line 39: nit: add a new line at the end of this file

.github/workflows/lint.yml
```diff
@@ -0,0 +1,19 @@
+name: Lint Code Base
```
ipmsteven, line 1: question: Will this block the PR or just show a warning when linter found some violation?

.github/workflows/lint.yml
```diff
@@ -0,0 +1,19 @@
...
+
+jobs:
+  build:
+    name: Lint Code Base
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Code
+        uses: actions/checkout@v2
+      - name: Lint Code Base
+        uses: docker://github/super-linter:v2.1.1
```
ipmsteven, line 16: could you check we are using the same bash ruleset which we already used in enterprise2?

.github/workflows/main.yml
ewgenius: Yeah, makes sense for now. But we will need to change it back, if will be adding older ubuntu versions in testing (once flaky test fixed for them)

.github/workflows/lint.yml
ewgenius: As far as I know `shellcheck` uses `#shellcheck` inline comments and `.shellcheckrc` file for single place configuration. We don't have such file in enterprise2, so I guess it falls back to inline directives and default configuration

.github/workflows/lint.yml
ewgenius: :+1:

ewgenius: 👍

.github/workflows/main.yml
ewgenius: 👍

.github/workflows/lint.yml
ewgenius: No, it won't:
- for now we have only Travis CI check as required. Once This PR completed, I'll switch it to build GH action, but linting can be optional

image

Issue #529, opened 2019-12-03
add time in output


ChristophHoyer:
it might be useful to see, how long a restore takes.
Therefore it would be helpful, if the ghe-restore command could print out a timestamp at the beginning and the end of the run.
Something like:

Checking for leaked keys in the backup snapshot that is being restored ...
* No leaked keys found
Connect github.superserver.com:122 OK (v2.18.7)

WARNING: All data on GitHub Enterprise appliance github.superserver.com (v2.18.7)
         will be overwritten with data from snapshot 20191203T011002.
Please verify that this is the correct restore host before continuing.
Type 'yes' to continue: yes

Starting restore of github.superserver.com:122 with backup-utils v2.18.0 from snapshot 20191203T011002
**Start time: 2019-12-03 11:30**
Stopping cron and github-timerd ...
Restoring UUID ...
Restoring MySQL database ...
Restoring Redis database ...
Restoring Git repositories ...
Restoring Gists ...
Warning: Gist backup missing. Skipping ...
Restoring GitHub Pages ...
Restoring SSH authorized keys ...
Restoring storage data ...
Restoring custom Git hooks ...
Restoring Elasticsearch indices ...
Restoring Audit logs ...
Restoring hookshot logs ...
Configuring appliance ...
Starting cron ...
Restoring SSH host keys ...
**End time: 2019-12-03 12:00
Runtime: 00:30h**
Restore of github.superserver.com:122 from snapshot 20191203T011002 finished.

bltstillwater:
Yes

cainejette:
@ChristophHoyer that's a great idea! I'm not sure we have bandwidth right now to add this, but we would happily review a PR if you submitted one!

djdefi:
It looks like this was implemented in https://github.com/github/backup-utils/pull/625

PR #625, opened 2020-08-01
Added basic timing around the ghe-restore process

johnclaus:
This PR adds basic timing output around the `ghe-restore` process, per the request from Issue #529

### DIFF
--- a/bin/ghe-restore
+++ b/bin/ghe-restore
@@ -174,6 +174,8 @@ if $instance_configured && ! $force; then
 fi
 
 # Log restore start message locally and in /var/log/syslog on remote instance
+START_TIME=$(date +%s)
+echo 'Start time:' $START_TIME
 echo "Starting restore of $GHE_HOSTNAME with backup-utils v$BACKUP_UTILS_VERSION from snapshot $GHE_RESTORE_SNAPSHOT"
 ghe_remote_logger "Starting restore from $(hostname) with backup-utils v$BACKUP_UTILS_VERSION / snapshot $GHE_RESTORE_SNAPSHOT ..."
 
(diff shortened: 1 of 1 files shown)


cainejette:
Awesome, thank you!

johnclaus:
@cainejette My PR is failing on linters erroring out on code I didn't change.  Is there any way of running these linters locally against my change?

cainejette:
> My PR is failing on linters

That shouldn't be a required step, so you should be clear from that perspective. Although we realize now that the action wasn't set up properly to trigger on PRs from forks. If you push a commit, will that trigger the two required build steps?

johnclaus:
> If you push a commit after that goes in, will that trigger the two required build steps?

Yes.  It seems to pass the [macos-latest](https://github.com/johnclaus/backup-utils/runs/946384416) step, but it fails on [ubuntu-latest](https://github.com/johnclaus/backup-utils/runs/946384405) and [Lint Code Base](https://github.com/johnclaus/backup-utils/runs/946384385).

The former error seems to be because of an outdated URI for downloading `shellcheck` and the latter appears to fail on Markdown linting errors unrelated to my PR.

cainejette:
> The former error seems to be because of an outdated URI for downloading `shellcheck`

This is also being fixed in #628 

> the latter appears to fail on Markdown linting errors unrelated to my PR.

This can be ignored

johnclaus:
@cainejette So can someone merge this or is there something else I need to do for this PR to go green?

cainejette:
> can someone merge this

Yep! We will see it through. Thanks for contributing!

cjonsmith:
Merged! Sorry for the delay, and thanks again for the contribution 😄

PR #627, opened 2020-08-03
Improve Python3 & Finding Moreutils Parallel 

robbat2:
The testcase w/ Python doesn't really need Python3, it just needs to validate python; so use `ast` instead of `compiler`, which is available in Py3.

Also support `moreutils` on other Linux distributions where the `parallel` binary might have a different location, e.g. Gentoo.

### DIFF
--- a/test/testlib.sh
+++ b/test/testlib.sh
@@ -431,3 +431,26 @@ verify_all_restored_data() {
   # verify common data
   verify_common_data
 }
+
+setup_moreutils_parallel() {
+  # CI servers may have moreutils parallel and GNU parallel installed.
+  # We need moreutils parallel
+  local x
+  for x in \
+      /usr/bin/parallel.moreutils \
+      /usr/bin/parallel_moreutils \
+      /usr/bin/moreutils.parallel \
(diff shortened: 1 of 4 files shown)


test/testlib.sh
```diff
@@ -431,3 +431,25 @@ verify_all_restored_data() {
...
+      /usr/bin/parallel_moreutils \
+      /usr/bin/moreutils.parallel \
+      /usr/bin/moreutils_parallel \
+      ; do
+        if [ -x "${x}" ]; then
+            ln -sf "${x}" "$ROOTDIR/test/bin/parallel"
+            break
+        fi
+  done
+  unset x
```
ryansimmen, line 448: `x` could be a local variable and then you wouldn't need to `unset` here.

robbat2: Fixed

ryansimmen:
@robbat2 please sync your fork with the latest master, merge master into your branch and push for the Actions tests to run in this PR.

robbat2:
@ryansimmen rebased on master now

PR #628, opened 2020-08-04
Use pull_request event for actions triggering

ewgenius:
- `push` event is not triggered by PR from forks
- updated shellcheck location

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,7 +1,7 @@
 name: Lint Code Base
 
 on:
-  pull_reqeust:
+  pull_request:
     branches-ignore:
       - 'master'
 
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,7 +1,7 @@
 name: Lint Code Base
 
 on:
-  push:
+  pull_reqeust:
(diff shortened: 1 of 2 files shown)


.github/workflows/lint.yml
```diff
@@ -1,7 +1,7 @@
 name: Lint Code Base
 
 on:
-  push:
+  pull_reqeust:
```
katie-simmons, line 4:
```suggestion
  pull_request:
```
a small typo correction

ewgenius: Thanks @katie-simmons! 👍

.github/workflows/lint.yml
```diff
@@ -1,7 +1,7 @@
 name: Lint Code Base
 
 on:
-  push:
+  pull_request:
```
cainejette, line 4: Does this mean it only happens once when opening the PR? What about pushing a commit to a branch with a PR open, would that rerun checks?

ewgenius: It is hard to tell that from GH actions docs, but it will queue action on any push to opened PR branch

I manually checked this behavior in separate repository: https://github.com/ewgenius/test-repository/pull/1

PR #639, opened 2020-09-18
Rename redis-cli to ghe-redis-cli

jlisam:
Similar to the one in the private repo, this PR replaces `redis-cli` for `ghe-redis-cli`.


zerowidth (approved):

cainejette (approved):
Approved assuming we can get the tests passing

jianghao0718:
I think it will broke back-compat in 2.20/2.21 if there is no ghe-redid-cli in release branches

zerowidth:
Updated for backward compatibility, and tested both in development and the current release.

PR #648, opened 2020-09-28
Building Docker image from git source

AhrazA:
We ran into issues attempting to match the version of backup-utils with the version of GitHub Enterprise we were running. This is because the Dockerfile always pulls the version tagged as "stable", regardless of which version of the repo is checked out.

By building from the git source files, instead of downloading the current release tagged "stable", users will be able to easily select which version of the backup-utils they wish to build.

As it stands, the only way to specify a version would be to roll a custom Dockerfile that either builds from source, or pulls in a specific release tag from `https://github.com/github/backup-utils/archive/`.


AhrazA:
Any updates regarding this?

.dockerignore
```diff
@@ -1,2 +0,0 @@
-*
-!share/github-backup-utils/ghe-docker-init
```
djdefi, line 2: I think we could just delete this entire file.

djdefi (commented):
Overall I think am ok with this approach, however two concerns come to mind:

1. The image size will be a bit larger, and builds will take a bit longer as Git history will get sent to the build context
2. This may change what folks using the Docker image are currently expecting to happen (i.e. run the `stable` branch).

cc: @github/backup-utils @github/ghes-infrastructure any thoughts?

AhrazA:
Thank you for the reply @djdefi 

I made some changes that address your first comment. Also applied the same patch to the alpine Dockerfile.

Any guidance on why the macOS test might be failing? I spent some time trying to figure it out to no avail.

AhrazA:
Pinging here again to keep this alive.

cainejette:
cc @github/backup-utils @github/ghes-infrastructure

djdefi:
> Any guidance on why the macOS test might be failing? I spent some time trying to figure it out to no avail.

I think this may be due to MacOS's default version of `stat` not supporting the same flags as the GNU Linux `stat`.

zachary-mark (approved):
Thank you for the contribution!

Issue #490, opened 2019-05-22
Operations CPU constrained on gzip


indygreg:
When looking at why our GHE backups were taking so long, I noticed that various operations are CPU constrained on gzip compression/decompression. For example, when dumping the MySQL database, I witnessed the MySQL process consume ~14 minutes of CPU time total and the `gzip` process it was piping to consuming ~45 minutes of CPU time! Put another way, if the compressor could ingest the line speed that MySQL dumping is capable of emitting, GHE backups would complete ~30 minutes faster on this instance on just the MySQL data bits alone. On the decompression side, the MySQL archive consumed ~39 minutes of CPU time with gzip.

Large backup operations could be substantially faster if a modern, faster compression library were used. I personally recommend zstd, which yields better *and* faster compression than zlib/gzip at default/normal compression levels. I zstd compressed the MySQL dump of our GHE instance (using level 3 - the default) and the resulting archive was ~75% the size of the gzip version and took far less CPU to compress. On the decompression side, it required ~130s of CPU versus ~2,400s.

In summary, various GHE backup operations are CPU constrained by gzip compression. Replacing gzip with something more modern like zstd will make these operations faster, substantially so on larger GHE instances.

shcorbett:
There have been a lot of improvements to overall performance for backup-utils. One area we do have in our backlog is to look at zstd (although initial tests did not show as significant of an improvement). Since we have this on our backlog to take a look at internally - going to close out this issue

PR #649, opened 2020-10-05
ISSUE-490: Replace gzip with zstd to improve CPU constraints

johnclaus:
This is a cursory pass at fixing the CPU constraints described in [ISSUE-490](https://github.com/github/backup-utils/issues/490) by swapping out calls to `gzip` with `zstd`.


tashaemery (approved):

johnclaus:
@tashaemery These tests are still failing?  Are they false positives?  If so, is there something I can fix to make these tests less flaky?

johnclaus:
@tashaemery I noticed that you approved this PR, but these tests are still failing.

djdefi (commented):
I believe this would also need to be added into the Dockerfiles ( https://github.com/github/backup-utils/blob/master/Dockerfile.alpine / https://github.com/github/backup-utils/blob/master/Dockerfile).

I'm not super familiar with `zstd`, so it would be good to make sure that it widely available in common distros (Debian, CentOs etc), and we'll definitely want to call it out as a [backup host requirement in the README](https://github.com/github/backup-utils/blob/master/docs/requirements.md#backup-host-requirements). 

cc: @github/ghes-infrastructure @github/backup-utils any thoughts as to needing to retain `gzip` as a fallback as well.

snh (changes requested):
Thank you for the PR @johnclaus!

Unfortunately we do not currently distribute `zstd` with GitHub Enterprise Server, and as most of these `zstd` calls are actually occurring on the GHES appliance, rather than the host running `backup-utils`, these will fail.

Additionally, `backup-utils` needs to be able to continue to restore older backup snapshots created using `gzip`, which I don't believe would be possible with this PR as it stands.

johnclaus:
@snh Thank you so much for your feedback and insight!  I'm assuming adding `zstd` to the GitHub Enterprise server requires an internal PR?  Let's say I change this PR to do the following:

1. Check if `zstd` is installed on the appliance.  If so, use `zstd` for all manners of compression.  Otherwise, default to `gzip`
2. Fix my logic around restoring old backup snapshots using `gzip`

Do you foresee any other reason for this PR not to land?

tashaemery (approved):

snh:
> Thank you so much for your feedback and insight! I'm assuming adding zstd to the GitHub Enterprise server requires an internal PR? Let's say I change this PR to do the following:

@johnclaus That's correct, I've passed this PR on to the Engineering team responsible, who will be able to offer more insight as to our roadmap here.

Regarding backwards compatibility, an approach we have used in the past is to use the remote GitHub Enterprise Server version to determine when to use a new method/approach.

If you call `ghe_remote_version_required` in the relevant `backup-utils` scripts, then `$GHE_REMOTE_VERSION`, along with `$GHE_VERSION_MAJOR`, `$GHE_VERSION_MINOR`, and `$GHE_VERSION_PATCH` will be automatically populated.

For example, say `zstd` landed in GitHub Enterprise Server 2.25.0:

```
ghe_remote_version_required "$GHE_HOSTNAME"

if [ "$GHE_VERSION_MAJOR" -eq 2 ] && [ "$GHE_VERSION_MINOR" -ge 25 ] || [ "$GHE_VERSION_MAJOR" -ge 3 ]; then
  compress_command="zstd"
else
  compress_command="gzip"
fi
```

And then use `$compress_command` in place of `gzip`/`zstd`.

snh:
Something else that came to mind along the line of what @djdefi raised is that introducing `zstd` as a dependency on the `backup-utils` host may not be something we want to do in the short term due to the wide variety of systems out there running `backup-utils`. Some of these systems are in highly controlled environments where installing an additional package can be a challenge.

With this in mind, it might be worth limiting the initial implementation of `zstd` to those cases where it is ran on the GHES appliance.

gamefiend:
Thanks for the PR @johnclaus and apologies for the delayed response! Because there are many dependencies and complicating factors regarding this replacement, we have decided to put this on our internal roadmap to work on. We will close this issue and look to do this work in a future release. 

Thanks again for your contribution!

PR #650, opened 2020-10-06
Skip config apply when restore settings

jianghao0718:
We can skip unnecessary config apply when restore settings, config apply will be later called in restore process.


omgitsads (approved):

jianghao0718:
This won't work, close this issue

PR #658, opened 2020-11-17
update usage of github-env-dispatch to support STDIN param

ipmsteven:
use [--cmd_meta](https://github.com/github/enterprise2/blob/9d2408144c3eb198c4c5767b999fe6495945a8a7/vm_files/usr/local/bin/github-env-dispatch#L33) to pass payload so that we could simulate a STDIN input for a command that is called through github-env-dispatch

slack conversation: https://github.slack.com/archives/CNDAXAKU1/p1605558662228300


jianghao0718:
Awesome! We need to do the same fix `ghe-restore-pages`, `ghe-restore-repositories-gist` and `ghe-restore-storage`

ipmsteven:
> Awesome! We need to do the same fix `ghe-restore-pages`, `ghe-restore-repositories-gist` and `ghe-restore-storage`

yea, I have pushed fix to support the rest of the resource type

jianghao0718 (approved):
LGTM

fremm102:
null

PR #659, opened 2020-11-17
stop github-timerd based on its running environment

ipmsteven:
resolves: https://github.com/github/ghes-infrastructure/issues/1397

### DIFF
--- a/bin/ghe-restore
+++ b/bin/ghe-restore
@@ -259,16 +259,28 @@ if $CLUSTER; then
     ghe_verbose "* Warning: Failed to stop cron on one or more nodes"
   fi
 
-  if ! ghe-ssh "$GHE_HOSTNAME" -- "ghe-cluster-each -- sudo service github-timerd stop"; then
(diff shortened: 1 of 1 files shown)


bin/ghe-restore
```diff
@@ -259,16 +259,28 @@ if $CLUSTER; then
     ghe_verbose "* Warning: Failed to stop cron on one or more nodes"
   fi
 
-  if ! ghe-ssh "$GHE_HOSTNAME" -- "ghe-cluster-each -- sudo service github-timerd stop"; then
-    ghe_verbose "* Warning: Failed to stop github-timerd on one or more nodes"
+  if ! ghe-ssh "$GHE_HOSTNAME" -- "ghe-cluster-each -- sudo systemctl list-units | grep github-timerd"; then
```
jianghao0718, line 262: Can we check `GHE_REMOTE_VERSION` instead? that will avoid one more ssh

jianghao0718 (approved):

PR #668, opened 2020-12-16
Add node version check and fail out if they don't match

ashishkeshan:
Recently ran into an issue where not all nodes were on the same version during a restore. Unfortunately, there were no checks in place, so the restore continued on and eventually failed much later without giving much of a hint for the reason.

### DIFF
--- a/bin/ghe-host-check
+++ b/bin/ghe-host-check
@@ -49,7 +49,7 @@ fi
 
 if "$CLUSTER"; then
   node_version_list=$(ghe-ssh "$host" ghe-cluster-each -- ghe-version)
-  echo "$node_version_list"
+  echo "$node_version_list" 1>&2
   first_node_version=""
   (
     IFS=$'\n'
--- a/bin/ghe-host-check
+++ b/bin/ghe-host-check
@@ -50,20 +50,11 @@ fi
 if "$CLUSTER"; then
   node_version_list=$(ghe-ssh "$host" ghe-cluster-each -- ghe-version)
   echo "$node_version_list" 1>&2
-  first_node_version=""
-  (
-    IFS=$'\n'
-    for node_version in $node_version_list; do
-      if [ -z "$first_node_version" ]; then
-        first_node_version=$(echo "$node_version" | cut -f2 -d":")
-      fi
-      current_node_version=$(echo "$node_version" | cut -f2 -d":")
-      if [ "$first_node_version" != "$current_node_version" ]; then
-        echo "Error: Not all nodes are running the same version! Please ensure all nodes are running the same version before using backup-utils." 1>&2
-        exit 1
-      fi
-    done
-  )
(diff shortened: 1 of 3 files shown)


bin/ghe-host-check
```diff
@@ -15,32 +15,57 @@ set -e
...
+      if [ -z "$first_node_version" ]; then
+        first_node_version=$(echo "$node_version" | cut -f2 -d":")
+      fi
+      current_node_version=$(echo "$node_version" | cut -f2 -d":")
+      if [ "$first_node_version" != "$current_node_version" ]; then
+        echo "Error: Not all nodes are running the same version! Please ensure all nodes are running the same version before using backup-utils." 1>&2
+        exit 1
+      fi
+    done
+  )
```
cainejette, line 66: A bash-ier option (which is not always preferable) could be something like this:

```suggestion
  distinct_versions=$(echo "$node_version_list" | awk '{split($0, a, ":"); print a[2]}' | uniq | wc -l)
  if [ $distinct_versions -ne 1 ]; then
    echo "Error: Not all nodes are running the same version! Please ensure all nodes are running the same version before using backup-utils." 1>&2
    exit 1
  fi
```

The `awk` command is splitting each line on `:` and storing the second half in an array named `a`

cainejette (approved):

ashishkeshan:
validated with my own cluster where in one run I kept the versions the same:
```
❯ bin/ghe-restore
Checking for leaked keys in the backup snapshot that is being restored ...
* No leaked keys found
build-cluster-app: GitHub Enterprise Server 3.1.0 lxc 2020-12-16 5e97c07602
build-cluster-data: GitHub Enterprise Server 3.1.0 lxc 2020-12-16 5e97c07602
build-cluster-util: GitHub Enterprise Server 3.1.0 lxc 2020-12-16 5e97c07602
build-cluster-bare: GitHub Enterprise Server 3.1.0 lxc 2020-12-16 5e97c07602
Connect 10.0.1.247:122 OK (v3.1.0)
WARNING: All data on GitHub Enterprise appliance 10.0.1.247 (v3.1.0)
         will be overwritten with data from snapshot 20201217T231023.
Please verify that this is the correct restore host before continuing.
Type 'yes' to continue: yes
Start time: 1608246794
Starting restore of 10.0.1.247:122 with backup-utils v3.1.0 from snapshot 20201217T231023
```
in another I manually edited the `etc/github/enterprise-release` to be an older version:
```
❯ bin/ghe-restore
Checking for leaked keys in the backup snapshot that is being restored ...
* No leaked keys found
build-cluster-app: GitHub Enterprise Server 2.19 lxc 2020-12-16 5e97c07602
build-cluster-data: GitHub Enterprise Server 3.1.0 lxc 2020-12-16 5e97c07602
build-cluster-util: GitHub Enterprise Server 3.1.0 lxc 2020-12-16 5e97c07602
build-cluster-bare: GitHub Enterprise Server 3.1.0 lxc 2020-12-16 5e97c07602
Error: Not all nodes are running the same version! Please ensure all nodes are running the same version before using backup-utils.
```

PR #673, opened 2021-01-13
Bump version: 3.0.0.rc

DangMH:
Includes general improvements & bug fixes and support for GitHub Enterprise Server v3.0.0.rc

/cc @github/backup-utils


share/github-backup-utils/version
```diff
@@ -1 +1 @@
-2.22.0
+3.0.0.rc
```
imjennyli, line 1: should we call this rc1 instead of rc?

PR #674, opened 2021-01-13
Bump version: 3.0.0.rc1

DangMH:
Includes general improvements & bug fixes and support for GitHub Enterprise Server v3.0.0.rc1

/cc @github/backup-utils


imjennyli (approved):

share/github-backup-utils/version
```diff
@@ -1 +1 @@
-2.22.0
+3.0.0.rc1
```
belaltaher8, line 1: im not sure if this should be `3.0.0.rc1` or just `3.0.0`. I believe this is just used to compare against the version of the instance that's being backed up so I don't think it has any concept of `gm`s or `rc`s.

test/testlib.sh
```diff
@@ -42,7 +42,7 @@ export GHE_BACKUP_CONFIG GHE_DATA_DIR GHE_REMOTE_DATA_DIR GHE_REMOTE_ROOT_DIR
 
 # The default remote appliance version. This may be set in the environment prior
 # to invoking tests to emulate a different remote vm version.
-: ${GHE_TEST_REMOTE_VERSION:=2.22.0}
+: ${GHE_TEST_REMOTE_VERSION:=3.0.0.rc1}
```
belaltaher8, line 45: same as above unsure if this should be `3.0.0.rc1` or just `3.0.0`

belaltaher8 (dismissed):

DangMH:
Current test failures are also present in `master` as is stands

https://github.com/github/backup-utils/runs/1704388764?check_suite_focus=true

This has also been noted in similar issues: https://github.com/github/backup-utils-private/issues/198

The test cases have also been manually verified and will be again after release.  These are considered test infrastructure failures and will be bypassed and resolved later

DangMH:
Issue for tracking test failures https://github.com/github/backup-utils/issues/677

imjennyli (approved):

PR #683, opened 2021-01-18
Fix restoring the password pepper for already configured instances

dbussink:
We normally skip settings restore when restoring into an already configured instance. This is to not overwrite / reset settings
unexpectedly. This is fine for all settings, except for the password pepper.

The password pepper is associated with the MySQL data and GitHub passwords used there, so it needs to be restored always together with the MySQL restore.

This moves the pepper restore to always be done together with the MySQL restore. We always here update the variable used here since the `restore-secret` function expects $GHE_RESTORE_SNAPSHOT_PATH to be set. We had a differently named variable in the MySQL restore with the same value, so that variable was renamed to match the `restore-secret` expectation so it can find the backed up password pepper.

### DIFF
--- a/share/github-backup-utils/ghe-restore-mysql
+++ b/share/github-backup-utils/ghe-restore-mysql
@@ -27,7 +27,7 @@ ghe_remote_version_required "$GHE_HOSTNAME"
 export GHE_RESTORE_SNAPSHOT
 
 # The directory holding the snapshot to restore
-snapshot_dir="$GHE_DATA_DIR/$GHE_RESTORE_SNAPSHOT"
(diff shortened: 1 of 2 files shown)


ryansimmen (approved):

amulyaraja:
nice, is there a backport for this coming soon @dbussink?

dbussink:
> nice, is there a backport for this coming soon @dbussink?

@amulyaraja I don't think a backport is needed here? 3.0.0 will afaik be tagged from `master`?

jianghao0718:
no backport is needed. We will release from master.

PR #686, opened 2021-01-20
Resolve "ghe-restore with host arg and config value" test

DangMH:
`ghe-ssh "$GHE_HOSTNAME" -- 'ghe-config app.actions.enabled'` always returns exit code 0 even if the actions config is disabled.

The fix resolves that by evaluating the config value then doing the appropriate comparison.


test/testlib.sh
```diff
@@ -525,7 +525,12 @@ enable_actions() {
 }
 
 is_actions_enabled() {
-  ghe-ssh "$GHE_HOSTNAME" -- 'ghe-config --true app.actions.enabled'
+  # ghe-ssh will always return exit code 0.  Evaluate property value instead.
+  if $( ghe-ssh "$GHE_HOSTNAME" -- 'ghe-config app.actions.enabled' ) = "true"; then
```
jianghao0718, line 529: 'ghe-config app.actions.enabled' is possible to return non zero exit code if this is not set. Probably a combination of `--present` and `--true` would help here

AliMousai (approved):

sonialey25:
> > - [ ] > > > []()[[___~~**_@@@@_**~~___]()]()

test/testlib.sh
sonialey25: there are hackers ising my info vell numbers and many account also be anle to change passwords  using Otiplex

q464pp:
可以拉取合併嗎謝謝

davidallen21151:
Get Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

q464pp:
請問可以拉取合併嗎謝謝

PR #719, opened 2021-04-09
Add backup cadence variable to the appliance

Samirat:
In order to check if backups are failing or are not being taken on the appliance, I want to get the variable from here to indicate how often they _should_ be getting taken.

### DIFF
--- a/share/github-backup-utils/ghe-backup-mssql
+++ b/share/github-backup-utils/ghe-backup-mssql
@@ -180,6 +180,9 @@ if [ -n "$backup_command" ]; then
   ghe-ssh "$GHE_HOSTNAME" -- "$backup_command" || failures="$failures mssql"
   bm_end "$(basename $0)"
 
+  # Configure the backup cadence on the appliance, which is used for diagnostics
+  ghe-ssh "$GHE_HOSTNAME" -- "ghe-config mssql.backup.cadence $cadence"
+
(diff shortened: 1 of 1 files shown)


share/github-backup-utils/ghe-backup-mssql
```diff
@@ -180,6 +180,9 @@ if [ -n "$backup_command" ]; then
   ghe-ssh "$GHE_HOSTNAME" -- "$backup_command" || failures="$failures mssql"
   bm_end "$(basename $0)"
 
+  # Configure the backup cadence on the appliance, which is used for diagnostics
+  ghe-ssh "$GHE_HOSTNAME" "ghe-config mssql.backup.cadence $GHE_MSSQL_BACKUP_CADENCE"
```
Samirat, line 184: Have to change the script to handle this format.

boxofyellow: So we consider making our live easier by change this to record the info we need.  The hours/days that back should happen.

Samirat: I'd rather have it be the same everywhere and parse it when we use it.

boxofyellow: Works for me.

boxofyellow (approved):

PR #720, opened 2021-04-13
Fix is_default_external_database_snapshot function

christopherb-stripe:
This function tries to access an argument ($1) but no invocations of this
function pass in an argument. The implementation has been changed to rely on
variables from the environment which allow it correctly find the appropriate
sentinel file.


bwestover (approved):
Looks great, thanks!

PR #726, opened 2021-05-07
Fix debian package versioning

henrich:
version number 3.0.1.rc1 is bigger than 3.0.1 official release,
so we should use ~ for it

$ dpkg --compare-versions 3.0.1.rc1 gt 3.0.1
$ echo $?
0

note: gt means "greater than"

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,4 +1,4 @@
-github-backup-utils (3.1.0.rc1) UNRELEASED; urgency=medium
+github-backup-utils (3.1.0~rc1) UNRELEASED; urgency=medium
 
   * Update repository backups to use ghe-gc-* for lock file management #188
   * A faster way to restore storage blobs (clusters) #212


zachary-mark (approved):
You are indeed correct, thank you for the contribution!

As far as I'm aware, this should only affect the versioning of the Debian packaging and not the version negotiation within GHES as that is performed elsewhere, so i'm pretty sure this is safe.

zachary-mark:
#727 should fix the broken required tests.  Once that merges I will merge master into this PR and get it merged down.

henrich:
@zachary-mark Can I ask some questions about backup-utils debian package?

I can improve more as an appropriate Debian package but I wonder there's any restriction and requirement for this tool. Usually, I made official Debian packages it complies with Debian releases, using git repo and gbp-buildpackage tool. However, it may be different for this.

* Which version/distribution is the target for the build?
* Is there any info about the build environment?

For example, now Debian moves debhelper v13 for packaging but this package uses old v9. However, if you are using old Debian/Ubuntu for build environment, it is what it is.

zachary-mark:
@henrich I'm a newer maintainer of backup-utils, so I can't speak for everyone, but I think the goal is to allow the packaging to build for as many distros as possbile.  backup-utils is architecture independent and relatively straightforward to package, and there isn't really a whole lot in it that would require us to use a new version of debhelper.  Based on what is in `debian/control`, it looks like we should be able to build this for jessie or later at this point.  My feeling is that we want to target *at least* `oldstable`.

henrich:
@zachary-mark Well, let me clarify a bit. 

>  My feeling is that we want to target at least oldstable.

Is "at least oldstable"  as a target "build" environment? Because if you'd build backup-utils .deb package on Debian unstable, its package can be installed in oldstable (and oldoldstable, and more) because there are no release-dependent dependencies.

Which version of the build environment should I satisfy?

zachary-mark:
@henrich I build it and use it myself on Stretch-based systems fairly often, so I think Stretch is still appropriate for now.

henrich:
@zachary-mark thanks, got it.

PR #730, opened 2021-05-11
Clarify moreutils-parallel requirement

maclarel:
On some *nix distributions, such as CentOS, `moreutils` is available through the package manager on the system (or through another repository like EPEL), however `moreutils-parallel` is provided as a package that must be installed separately.

If only `moreutils` is installed on CentOS, for example, Backup Utilities will still fail to run and complain that `parallel` cannot be found despite the `moreutils` package being installed.

This Pull Request clarifies that `moreutils-parallel` may need to be installed separately depending on platform/distribution in order to help address the confusion that may arise from the lack of information on the `moreutils` website that is linked in documentation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,7 +3,7 @@
 This repository includes backup and recovery utilities for
 [GitHub Enterprise Server][1].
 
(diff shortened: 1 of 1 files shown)


gpadmesh:
+1

PR #733, opened 2021-05-17
Fix redirection bug (SC2259) in test/testlib.sh

henrich:
ShellCheck 0.7.2 says
> In test/testlib.sh line 149:
>    grep -a -v -e '^\+ end_test' -e '^+ set +x' <"$TRASHDIR/out" |
>                                                ^-- SC2259: This redirection overrides piped input. To use both, merge or pass filenames.
>
>For more information:
>  https://www.shellcheck.net/wiki/SC2259 -- This redirection overrides piped ...

so specify "$TRASHDIR/out" as the filename


bwestover (approved):
Thanks!

PR #732, opened 2021-05-13
Fix and make its Debian package modern

henrich:
With the power of lintian, lintian-brush, and my DD experience, fix all lintian warnings for the deb package (bonus: add man pages).

And also fix this package as arch-indep package (arch: amd64 -> arch: all). It means we can install this package on arm64 hosts without a rebuild.

PR #773, opened 2021-09-09
change amd64.deb to all.deb

jianghao0718:
# Context
During 3.2.0 release we found that due to change in https://github.com/github/backup-utils/pull/732, deb file suffix should be changed from  `amd64.deb` to `all.deb` as architecture is changed to `all`, this changes the deb file we ship as release.


zachary-mark (approved):

bwestover (approved):

hamedalex1:
#787

PR #770, opened 2021-09-03
Clear GitHub Connect settings when not restoring settings

lildude:
At the moment, when a backup it restored _without_ any settings, ie without using the `--config` flag, the GitHub Connect settings are still restored. This is because they're stored in the MySQL database and there's nothing in place to clear them or prevent them from being restored.

This leads to a problem where the restored instance will have all the same settings and GitHub Connect-specific unique ID as the original host and will start competing with the original instance if both are active. This can result in one instance being kicked off GitHub Connect, or confusing license sync information, if both remain connected.

This PR resolves this problem by calling a new script immediately after the MySQL database has been restored. This script will be added to all currently supported versions of GHES.

PR #775, opened 2021-09-10
Revert "Clear GitHub Connect settings when not restoring settings"

jianghao0718:
Reverts github/backup-utils#770

Other than revert, do we want to call `ghe-reset-gh-connect-server-id` instead?


lildude:
Whoops. The error is in the `else`. I can fix this on Tuesday.

jianghao0718:
Close in favor of a fix from @deleonjavier

PR #779, opened 2021-09-13
fix bug in is_service_external

jianghao0718:
Seems `is_service_external` is not returning correctly for external database when parsing `settings.json`

### DIFF
--- a/share/github-backup-utils/ghe-backup-config
+++ b/share/github-backup-utils/ghe-backup-config
@@ -417,7 +417,7 @@ is_service_external(){
     "mysql")
       if [ -n "$config_file" ]; then
         enabled=$(GIT_CONFIG="$config_file" git config mysql.external.enabled)
-        [ "$enabled" ];
+        [ "$enabled" = true ];
       else
         ghe-ssh "$GHE_HOSTNAME" -- ghe-config --true "mysql.external.enabled"
       fi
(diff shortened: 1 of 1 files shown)


cjonsmith (changes requested):

share/github-backup-utils/ghe-backup-config
```diff
@@ -417,7 +417,7 @@ is_service_external(){
     "mysql")
       if [ -n "$config_file" ]; then
         enabled=$(GIT_CONFIG="$config_file" git config mysql.external.enabled)
-        [ "$enabled" == "true" ];
```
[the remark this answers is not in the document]
jianghao0718: updated

share/github-backup-utils/ghe-backup-config
```diff
@@ -417,7 +417,7 @@ is_service_external(){
...
+        [ "$enabled" = true ];
```
cjonsmith, line 420: I think this change is functionally the same as what we already have; `=` is just the POSIX compliant version of `==` and the quotes aren't doing much on the right hand operator here since it's only a single word.

cjonsmith (changes requested):

share/github-backup-utils/ghe-backup-config
jianghao0718: during my testing, I got following if `$enabled=true`
```
zsh: = not found
```

PR #800, opened 2021-10-19
Fix: ghe-restore add force option when restarting actions

coralinepetit:
# Why
The attempt to start github actions while the github instance is in maintenance mode, cause the `ghe-restore` to exit with code 1.

The restart of github action sis caused by the check on the instance`--true app.actions.enabled` forcing the restart if actions are enables

# How
Add `-f` option to force the reload of github actions while the instance is in maintenance mode. Allow the `bin/ghe-restore` to run with an exit code 0 !


Claveyo1:
No

PR #815, opened 2021-11-17
Always default port to 122

jianghao0718:
## Context
For Github enterprise server the SSH port is always exposed as 122. In `backup-utils`, there are [some logics](https://github.com/github/backup-utils/blob/d792b23737f5e026d76a27f08c7523202c393506/bin/ghe-host-check#L55-L81) unnecessarily check connection ports and default to 22 first if not set and then it will failback to 122 after expected unsuccessful connection. Instead, we just default to 122. If port is set incorrect in `backup.config`, `backup-utils` should bail out.

### DIFF
--- a/bin/ghe-host-check
+++ b/bin/ghe-host-check
@@ -52,37 +52,8 @@ options="
 port=$(ssh_port_part "$host")
 hostname=$(ssh_host_part "$host")
 
-set +e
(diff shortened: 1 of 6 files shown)


qnordic (approved):
Much improved logic to ensure we are utilizing port 122

lildude (approved):
Makes sense to me.

PR #820, opened 2021-11-23
Skip proxy for backup / restore ssh config

jianghao0718:
In this PR we updated `ghe-ssh-config` to generate a ssh config by default skip proxy using `GHE_HOSTNAME`. This ssh config is used by backup / restore for repo / storage / pages which are traffic-heavy operation which might trash the proxy node, especially when backup is huge.

*Notes*
- This should impact cluster backup / restore only.
- We assume backup host has network access to ALL nodes in cluster.

Thanks a lot for hints from @snh .


rakira2309 (approved):

rakira2309 (approved):

qnordic (approved):
This looks good to me and with the recent testing, I think we're good to get it merged.

PR #804, opened 2021-11-06
fixed the problem that memory overflows when there are too many files…

Kiyokazu-g:
im reporting this issue in support ticket( #1380328 ).

---
Fix such an error:

```
Restoring storage data ...
/data/user/_ghe_backup/backup-utils/share/github-backup-utils/ghe-restore-storage: xrealloc: cannot allocate 18446744072682251136 bytes
````

PR #834, opened 2021-12-10
Revert "fixed the problem that memory overflows when there are too many files…"

jianghao0718:
Reverts github/backup-utils#804

This PR mentioned above might broke `ghe-restore-storage`, testing WIP

### DIFF
--- a/share/github-backup-utils/ghe-restore-storage
+++ b/share/github-backup-utils/ghe-restore-storage
@@ -26,6 +26,15 @@ GHE_HOSTNAME="$1"
 # us run this script directly.
 : ${GHE_RESTORE_SNAPSHOT:=current}
 
+# Find the objects to restore
+storage_paths=$(cd $GHE_DATA_DIR/$GHE_RESTORE_SNAPSHOT/ && find storage -mindepth 4 -maxdepth 4 -type f -exec wc -c {} \;)
+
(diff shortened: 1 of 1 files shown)


jianghao0718:
cc @domofactor / @Kiyokazu-g  -- I reverted the change as it broke our tests for restoring storage. I think this needs some work before we can merge.

Kiyokazu-g:
hi @jianghao0718 and @domofactor 

what kind of error did you get?
i would like to fix the current way because it fails when there are too many files.

jianghao0718:
In the test, we have seen
```
    Restoring storage data ...
    * Transferring data to 10.0.1.239 ...
    rsync: link_stat "/workspace/enterprise2/repos-buster/backup-utils/data/20211210T200511/storage/0/0/0" failed: No such file or directory (2)
    rsync error: some files/attrs were not transferred (see previous errors) (code 23) at main.c(1196) [sender=3.1.2]
    sending incremental file list
    
    sent 22 bytes  received 12 bytes  68.00 bytes/sec
    total size is 0  speedup is 0.00
    ++<test-backup.sh:174 test-backup.sh:1 restore_from_backup()> kill 15759
    +<test-backup.sh:176 lib.sh:307 end_test()> test_status=23
```

Issue #797, opened 2021-10-07
Fix: exit code status when ghe-restore try to start ghe-actions while in maintenance mode


GuyMauve:
The `ghe-restore` script try to run the `ghe-actions-start` when github actions are enabled. (added in PR #769)
On the server, the call of `ghe-actions-start` cause the following output: `Maintenance mode is set. Skip starting Actions.` For unknown reasons, this results on an `exit 1` code from the `ghe-restore` script. 
This behavior is bypassed by adding the `-f` option to the `ghe-actions-start` (line 96 of the [ghe-restore script](https://github.com/github/backup-utils/blob/master/bin/ghe-restore#L96)) !

rdelany:
Fixed with https://github.com/github/backup-utils/pull/836

PR #836, opened 2021-12-17
Fix compat issue with ghe-actions-start during maintenance mode

rdelany:
Issue: https://github.com/github/backup-utils/issues/797

The problem occurs when using a newer version of backup-utils (3.2+ that includes [this change](https://github.com/github/backup-utils/pull/769/files)) with an older version of GHES (pre-3.3 so it doesn't include [this change](https://github.com/github/enterprise2/pull/26845)) .

We get the following error:
```
Restarting Actions after restore ...
Maintenance mode is set. Skip starting Actions.
```

This change makes it so we use `ghe-actions-start -f` in pre-3.3 builds so that it starts and succeeds even in maintenance mode, and `ghe-actions-start` in 3.3+ builds. We can't use `ghe-actions-start -f` everywhere like the [initial PR](https://github.com/GuyMauve/backup-utils/pull/1) for this issue, since in 3.3+ builds, the `-f` option has been removed and it will fail like so:
```
admin@rdelany-r31isulu4-service-bpdev-us-east-1-github-net-primary:~$ ghe-actions-start -f
Unrecognized argument: -f
...
```

### DIFF
--- a/bin/ghe-restore
+++ b/bin/ghe-restore
@@ -93,7 +93,14 @@ cleanup () {
 
   if ghe-ssh "$GHE_HOSTNAME" -- 'ghe-config --true app.actions.enabled'; then
     echo "Restarting Actions after restore ..."
-    ghe-ssh "$GHE_HOSTNAME" -- 'ghe-actions-start' 1>&3
+    # In GHES 3.3+, ghe-actions-start no longer has a -f (force) flag. In GHES 3.2 and below, we must provide the
+    # force flag to make sure it can start in maintenance mode. Use it conditionally based on whether it exists
+    # in the --help output
+    if ghe-ssh "$GHE_HOSTNAME" -- 'ghe-actions-start --help' | grep -q force ; then
+      ghe-ssh "$GHE_HOSTNAME" -- 'ghe-actions-start -f' 1>&3
+    else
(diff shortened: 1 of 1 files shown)


ritchxu (approved):

Rogergarc78 (commented):
Y

ASISBusiness:
> Issue: #797
> 
> The problem occurs when using a newer version of backup-utils (3.2+ that includes [this change](https://github.com/github/backup-utils/pull/769/files)) with an older version of GHES (pre-3.3 so it doesn't include [this change](https://github.com/github/enterprise2/pull/26845)) .
> 
> We get the following error:
> 
> ```
> Restarting Actions after restore ...
> Maintenance mode is set. Skip starting Actions.
> ```
> 
> This change makes it so we use `ghe-actions-start -f` in pre-3.3 builds so that it starts and succeeds even in maintenance mode, and `ghe-actions-start` in 3.3+ builds. We can't use `ghe-actions-start -f` everywhere like the [initial PR](https://github.com/GuyMauve/backup-utils/pull/1) for this issue, since in 3.3+ builds, the `-f` option has been removed and it will fail like so:
> 
> ```
> admin@rdelany-r31isulu4-service-bpdev-us-east-1-github-net-primary:~$ ghe-actions-start -f
> Unrecognized argument: -f
> ...
> ```

PR #838, opened 2021-12-23
Take full mssql backup if Actions backup job has interfered

zarenner:
Actions now has an internal mechanism named `TransactionLogMaintenanceJob` which tries to detect if backup-utils is not running, and if so takes its own backups to prevent runaway transaction log growth.

In rare cases where this job does take a backup and then backup-utils runs again (e.g. `backup-utils` temporarily turned off), the diff or transaction log backup may have the wrong full backup base. That is, the new diff/transaction backup may be based on the full backup that `TransactionLogMaintenanceJob` took rather than the last backup-utils full backup. In such a case, the set of backup-utils files does not represent a complete LSN chain and may not be restorable.

This PR attempts to fix this issue by detecting if the internal Actions job run since the last completed backup-utils full backup. If so (and a diff/transaction backup is due to run) the backup will be "upgraded" to a full backup to reset the chain and ensure the backup-utils snapshot is restorable.

### DIFF
--- a/share/github-backup-utils/ghe-backup-mssql
+++ b/share/github-backup-utils/ghe-backup-mssql
@@ -14,8 +14,7 @@ set -e
 backup_dir="$GHE_SNAPSHOT_DIR/mssql"
 last_mssql=
 backup_command=
-take_full=
-take_diff=
+backup_type=
 full_expire=
 diff_expire=
 tran_expire=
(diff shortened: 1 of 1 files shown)


share/github-backup-utils/ghe-backup-mssql
```diff
@@ -90,13 +89,69 @@ ensure_same_dbs() {
   fi
 }
 
+no_conflicting_backups_exist_db() {
+  backups_dir=$1
+  db=$2
+  ghe_verbose "Checking if the backup job has ran for '$db' since the last backup-utils backup..."
+
+  latest_full_backup=$(find "$backups_dir" -type f -name "$db*.bak" | egrep '[0-9]{8}T[0-9]{6}' | sort | tail -n 1)
+  latest_full_backup_file="${latest_full_backup##*/}"
```
zarenner, line 98: I'm currently testing each database _individually_ based on its latest backup file.

Alternatively, instead of checking each database individually I could instead find the oldest (by timestamp) full backup for _any_ database, and search everything newer. This would be faster and is probably fine since we always take backups of all DBs together... but it still seems a bit odd to me and might make future improvements like checking LSN chains harder.

@boxofyellow thoughts?

boxofyellow: This gets my vote.

boxofyellow (commented):
The changes look good to me.

share/github-backup-utils/ghe-backup-mssql
zarenner: Per discussion we just had, will keep current behavior of individually inspecting each database. This seems like a safer approach since we want to exclude checking other databases that shouldn't be backed up (e.g. unused partition dbs or non-actions dbs). It also makes it easier to add additional checks (e.g. LSN chain) in the future.

zarenner:
@boxofyellow After discussion with @rdelany I've decided to abandon this PR in favor of a new PR https://github.com/github/backup-utils/pull/843, which checks that LSN values are correct before taking diff/transaction backups.

We went through a number of customer issues which would not have been resolved by this PR, because it wasn't `TransactionLogMaintenanceJob` interfering; it was the customer's additional backup solutions (e.g. running multiple copies of backup-utils).

q464pp:
可以拉取合併嗎謝謝

PR #839, opened 2021-12-27
Fix broken macOS tests (except parallel restore)

zarenner:
Fixes long-broken macOS test run by:
1. Skipping `brew unlink parallel`, which is failing and now appears to be unnecessary
2. Adding a missing test stub for `systemctl`
3. Temporarily disabling parallel restore tests, which have remaining rsync issues documented in https://github.com/github/backup-utils/issues/841

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -21,7 +21,6 @@ jobs:
     - name: Install Dependencies (macOS)
       run: |
         brew install gnu-tar shellcheck jq pigz coreutils gnu-sed gnu-getopt
-        brew unlink parallel
         brew install moreutils gawk
       if: matrix.os == 'macos-latest'
     - name: Get Sources
--- a/test/bin/systemctl
+++ b/test/bin/systemctl
@@ -1,0 +1,1 @@
+ghe-fake-true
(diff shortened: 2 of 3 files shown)


share/github-backup-utils/ghe-backup-storage
```diff
@@ -117,7 +117,8 @@ fi
 bm_start "$(basename $0) - Storage object sync"
 for file_list in $tempdir/*.rsync; do
   hostname=$(basename $file_list .rsync)
-  storage_user=$(ghe-ssh $ssh_config_file_opt $hostname:$port -- stat -c %U /data/user/storage || echo git)
```
zarenner, line 120: `stat` has different format options between GNU and BSD. Use ls+awk which works across OSes. Alternative considered: detect macos and use `gstat` from brew-installed coreutils.

zarenner:
One of the linux test runs has some flakiness still 🙁... but can't see how it'd be related to this PR.

jianghao0718 (approved):

jianghao0718:
Thanks for fixing this! Sorry we missed the notification.

PR #843, opened 2021-12-29
Take full mssql backup if other backup mechanisms have interfered

zarenner:
This PR attempts to fix a couple different scenarios where customers have had unrestorable mssql backups taken, mainly due to breaks in chain of transaction log backups. These breaks are caused by:

1. Customers running multiple instances of backup-utils, or taking additional backups outside of backup-utils. These backups interfere as they "move up" the next log LSN that backup-utils will take, leaving a gap between backup-utils log backups. This can also occur with diff backups, where the "differential base" (the full backup the diff is based on) is unexpectedly changed to one that backup-utils doesn't actually have. 
2. Actions now has an internal mechanism named TransactionLogMaintenanceJob which tries to detect if backup-utils is not running, and if so takes its own backups to prevent runaway transaction log growth. In rare cases where this job does take a backup and then backup-utils runs again (e.g. backup-utils temporarily turned off due to an issue), the next backup-utils diff or log backup may have an incorrect base or transaction log gap.

We attempt to solve this by running additional checks before a diff or transaction backup is taken, to ensure that:
1. We have the correct full backup file for the diff we're about to take. We look up the LSN for our full backup file, and check it against what SQL says is the base that will be used for the next diff backup we're about to take.
2. Our most recent log file ends at the LSN that SQL says the next log backup will start at. Note that since our full backups always also include a compatible* log backup, when taking a log backup we don't need to check it against the full backup, just against the most recent log backup.

\* I believe there will still be rare scenarios where a race could result in a unrestorable backup. Specifically, because `ghe-export-mssql` additionally performs a log backup after a full/diff backup, if an external backup mechanism interfered during that short timeframe a gap could still form between the full/diff backup and log backup. If we decide to address this we should probably fix it in `ghe-export-mssql` itself. In the meantime, improved docs and customer education not to run multiple mssql backup mechanisms should also help.


share/github-backup-utils/ghe-backup-mssql
```diff
@@ -128,25 +181,89 @@ else
...
+      # Ensure that a diff backup will be based on the full backup file we have (rather than one another backup mechanism took)
+      if [ "$backup_type" == 'diff' ]; then
+        full_backup_file=$(get_latest_backup_file "$last_mssql" "$db" "bak")
+        if [[ "$full_backup_file" == "" ]]; then
+          ghe_verbose "Taking a full backup instead of a diff backup because for $db a full backup file wasn't found"
+          backup_type="full"
+          break
+        fi
+
+        full_backup_file_checkpoint_lsn=$(get_backup_first_lsn "$db" "$full_backup_file")
```
zarenner, line 208: fix

boxofyellow (approved):
LGTM: :shipit:

Do you think we should add some steps to the GHES test manifest to run though these checks with a real SQL back end?

zarenner:
@boxofyellow I'll create an internal issue for those manifests.

q464pp:
可以拉取合併嗎謝謝

zarenner:
@dhadka @ritchxu any concerns on this change (e.g. scenarios that might interfere / change LSNs)? I still need to test with HA, but since we're not looking through arbitrary entries in `msdb.dbo.backupset` to determine continuity I don't think the extra rows we saw there (from seeding operations?) will conflict.

share/github-backup-utils/ghe-backup-mssql
```diff
@@ -128,25 +181,89 @@ else
...
+        break
+      fi
+
+      # The starting LSN of the backup we're about to take must be equal to (or before) the last LSN from the last backup,
+      # otherwise there'll be a gap and the logfiles won't be restorable
+      if [[ "$next_log_backup_starting_lsn" -gt "$latest_log_backup_last_lsn" ]]; then
+        ghe_verbose "Taking a full backup instead of a $backup_type backup because for $db a gap would exist between the last backup ending at LSN $latest_log_backup_last_lsn and next backup starting at $next_log_backup_starting_lsn"
+        backup_type="full"
+        break
+      fi
```
ritchxu, line 258: Does this section need to be wrapped behind `if [ "$backup_type" == 'transaction' ]`

ritchxu (approved):
Seems reasonable to me! Left one small comment/question though.

Regarding the race condition comment in PR description, would it help if we add a lock file mechanism to disallow parallel run of backup-utils? The lock file could contain backup server watermark so if we detect it differs from where it's running, we know to take a full backup too.

share/github-backup-utils/ghe-backup-mssql
zarenner: Good question. The idea was to also upgrade diff backups to full (this is under `if [ "$backup_type" == 'diff' ] || [ "$backup_type" == 'transaction' ]`) since they also take a log backup, but you're quite possibly right that it isn't necessary. Let's think it through:

To restore to an arbitrary point before the diff backup, you need the full backup and log backups going from the full backup LSN up to the diff backup's LSN. The log backup chain should be intact (because of this check for log backups), _except_ possibly right before the diff backup if another mechanism took a log backup first.

To restore to a point at/after the diff backup, you need the full backup, diff backup, and only logs starting from at/before the diff backup's LSN. The log backup we take right after the diff backup will always cover its LSN, unless the unlikely race discussed elsewhere occurs.

So, I think that checking log continuity when taking a diff (+ log) backup doesn't actually increase the available restore points, but does guarantee that the set of backup files in a backup-utils snapshot can be PITR'ed to _any_ point (and "resets" the set of files to just be a new full backup).

Am I thinking about that correctly?

zarenner:
@ritchxu a lockfile would certainly would help for multiple backup-utils running. I _think_ that's the most common problem happening and as such would be a reasonable solution, but of course wouldn't solve it for ad-hoc backups / 3rd party backup solutions.

share/github-backup-utils/ghe-backup-mssql
ritchxu: > (this is under if [ "$backup_type" == 'diff' ] || [ "$backup_type" == 'transaction' ]) since they also take a log backup

I think this (i.e. diff backup also takes a log backup) answers my initial question 😃

ritchxu:
> but of course wouldn't solve it for ad-hoc backups / 3rd party backup solutions.

Agreed, I didn't know customers would use solutions other than backup-utils to perform the task 😱 . But your solution would handle that well so 👍🏼

zarenner:
Tested that HA automatic seeding backups do not conflict with this detection. Merging.

PR #866, opened 2022-02-09
Added support for Artifact cache service

tiwarishub:
This PR is adding backup utils for AC service. Backup and restore doesn't require much change apart from adding Repair database login for AC service (similar to token, mps and actions). 
As this change needs to be backwards compatible,  it is checking whether ac is there in GHES or not, then only it will run command for ac service. 

Testing scenario executed:
Testing was done using a single gheboot instance.
Test1:
Gheboot with ac enabled https://tiwarishub-005817a450c4dae53.gh-quality.net (on backup and restore l)
 Added repo ghe-admin/test 
 Create one workflow cache.yml (using actions/cache). Ran save/restore scenario
 Took backup
 Create one more workflow cache2.yml. Ran save/restore sceanrio
 Create one more repo ghe-admin/test3 and creted and ran wf cache.yml in it
 Put instance in maintenance
 Restore the instance to the place where the backup was taken.

Test2:
 Gheboot (3.3) without ac enabled http://tiwarishub-0c0ed174e544ec531.gh-quality.net  (on backup and restore)
 Added repo ghe-admin/test 
 Create one workflow
 Take backup
 Create one more repo ghe-admin/test2
 Put instance in maintenance
 Restore the instance to the place where the backup was taken


boxofyellow (approved):
LGTM: :shipit:

This PR made me think of some other cases that we will want to follow up on, but since they are outside the scope of this PR, I'll follow up offline.

PR #867, opened 2022-02-10
Minor Update of usage.md to include missing step

kerryhatcher:
If you don't put the target into maintenance the restore command will fail informing the user to put the target into maintenance.


DONALDSULLENDER (approved):

bonsohi (approved):
Good to merge! 👍

PR #878, opened 2022-03-14
Added support for AC databases

tiwarishub:
This PR adds support for AC service for backup and restore.
No changes are required for the backup scenario but for restore, we need a few changes.
1. Apply database repair login of AC databases, if snapshot contains AC databases.
2. Exiting the restore script, if appliance contains AC service but snapshot does not contain AC data. Reason for adding this I have explained in detail [here](https://github.com/github/c2c-actions/issues/3307#issuecomment-1064562924). Please check


bin/ghe-restore
```diff
@@ -262,8 +262,17 @@ if $instance_configured; then
...
 # Make sure the GitHub appliance has Actions enabled if the snapshot contains Actions data.
+# If above is true, also check if ac is present in appliance then snapshot should also contains ac databases
 if [ -d "$GHE_RESTORE_SNAPSHOT_PATH/mssql" ] || [ -d "$GHE_RESTORE_SNAPSHOT_PATH/actions" ]; then
-  if ! ghe-ssh "$GHE_HOSTNAME" -- 'ghe-config --true app.actions.enabled'; then
+  if ghe-ssh "$GHE_HOSTNAME" -- 'ghe-config --true app.actions.enabled'; then
+    ac_db_ghe=$(echo 'ghe-mssql-console -y -n -q "SELECT name FROM sys.databases" | grep -i "ArtifactCache" | wc -l | tr -d " "' | ghe-ssh "$GHE_HOSTNAME" /bin/bash)
+    ac_db_snapshot=$(find "$GHE_DATA_DIR/$GHE_RESTORE_SNAPSHOT/mssql/" -maxdepth 1 -name 'ArtifactCache*.bak' | wc -l | tr -d " ")
+    if [[ $ac_db_ghe -gt 0  &&  $ac_db_snapshot -eq 0 ]]; then
+      echo "Error: $GHE_HOSTNAME contains ArtifactCache databases but no ArtifactCache databases are present in snapshot. Aborting" 1>&2
+      echo "Please delete ArtifactCache databases from $GHE_HOSTNAME and retry" 1>&2
```
boxofyellow, line 272: I think this needs a link to a public doc.

Basically as a customer if I get an instruction to delete a database the very first thing they will do is open a support ticket.  By providing a link to a public doc that will give them the option to self server.

To should tell them the following
- How to confirm which databases to delete
- What specific commands should be used to perform the delete.

tiwarishub: I have added a public doc link (PR: https://github.com/github/docs-internal/pull/26254) in the error message. This will be public once GHES 3.5 is released and will require some guidance for PR review from the TPM side as well. But the instruction we have called out in the doc.

boxofyellow: Thanks!

Issue #560, opened 2020-02-11
Backups can fail due to "grep: write error: Resource temporarily unavailable"


maclarel:
**Issue:**

Backups can intermittently fail due to `grep: write error: Resource temporarily unavailable`. This appears to be linked to a significant (hundreds of thousands or more) Pages objects in an environment causing the `grep` process used in conjunction with output redirection to run out of resources and fail.

**Workarounds:**

[Remove the `egrep`s from `ghe-rsync`](https://github.com/github/backup-utils/blob/master/share/github-backup-utils/ghe-rsync#L28-L30)

**Potential fix:**

In combination with potential changes to output filtering (removal of `egrep`s), investigate additional logic to backup Pages in a similar fashion to repository networks to lower the number of files being transferred.

shcorbett:
It looks like [this PR](https://github.com/github/backup-utils/pull/881) may have resolved this but confirming if this can be closed?

shcorbett:
Will go ahead and close this out - please reopen if issue still exists and not covered with PR above. Thanks.

PR #881, opened 2022-03-22
simplify complex redirects for ghe-rsync

domofactor:
This updates `ghe-rsync` to simplify the redirect logic that was required in order to support both rsync `< 3.x` and `>= 3.x`. 

because we're already redirecting stderr into stdout when sending to the logs like the [example docs](https://github.com/github/backup-utils/blob/master/docs/scheduling-backups.md#example-scheduling-usage), users shouldn't see any difference with this change.

cc/ https://github.com/github/ghes/issues/3284, https://github.com/github/backup-utils/issues/560


manue1 (approved):

donal (approved):
Looks good — acknowledging that the change to redirect stderr to stdout (for 3.x+) is deliberate.

PR #886, opened 2022-04-05
Handling disabled Actions Cache service during restore

aparna-ravindra:
Following the PR in [enterprise2 repo ](https://github.com/github/enterprise2/pull/28668), ArtifactCache service in Actions is introduced as an optional service in GHES 3.5

Therefore, the restore script in the backup-utils will need to handle the case where there is a mismatch in the state of the target instance and the snapshot being restored. 

This PR introduces a check to confirm that the state of the ArtifactCache service matches in the target instance and the snapshot, and provides steps to the admin to bring them in the same state before snapshot restoration.

This is a followup PR to the previous https://github.com/github/backup-utils/pull/878


tiwarishub (approved):

boxofyellow (approved):
LGTM :shipit:

PR #894, opened 2022-04-12
Backport 3.3.2 restart cron on restore failure

donal:
Manual backport of https://github.com/github/backup-utils/pull/883 to a release candidate for 3.3.2.

PR #896, opened 2022-04-12
On restore failure, restart cron on target host

donal:
Manual backport of https://github.com/github/backup-utils/pull/883 to a release candidate for 3.3.2.

Corrects https://github.com/github/backup-utils/pull/894.


donal:
We can ignore the linter errors — that runs on PRs but not when the base branch is `master` (ie. we ignore it).

PR #903, opened 2022-04-21
On restore failure, restart cron on target host

donal:
Manual backport of https://github.com/github/backup-utils/pull/883 to a release candidate for 3.3.2.

### DIFF
--- a/bin/ghe-restore
+++ b/bin/ghe-restore
@@ -86,6 +86,19 @@ while true; do
   esac
 done
 
+start_cron () {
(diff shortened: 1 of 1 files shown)


donal:
We can ignore the linter errors — that runs on PRs but not when the base branch is `master` (ie. we ignore it).

donal:
macos were failing for a long time and not fixed until 3.4.0: https://github.com/github/backup-utils/pull/839 (ie. we can ignore the failure).

PR #911, opened 2022-05-03
Fix spelling mistake on requirements.md

mcfluffinstein:
I noticed a small spelling mistake when sending this documentation out to a customer. It's only minor, but I thought it would be worth fixing.


bonsohi (approved):
Good to merge when CI green! 👍

PR #923, opened 2022-05-26
Bump the hardcoded check values in the host check test to unbreak CI

jsoref:
https://github.com/github/backup-utils/pull/913#issuecomment-1139016893


jsoref:
@SteveCulver

elee (approved):
Thanks for the fix @jsoref

PR #924, opened 2022-05-27
Updates test-ghe-host-check

alejndr0:
Update test-ghe-host-check.sh to verify support against latest.json and remove hardcoded tests

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -13,7 +13,7 @@ jobs:
     - name: Install Dependencies (Linux)
       run: |
         sudo apt-get update -y
-        sudo apt-get install -y devscripts debhelper moreutils fakeroot jq pigz help2man wget
+        sudo apt-get install -y devscripts debhelper moreutils fakeroot jq pigz help2man
         wget "https://github.com/koalaman/shellcheck/releases/download/latest/shellcheck-latest.linux.x86_64.tar.xz"
         tar --xz -xvf "shellcheck-latest.linux.x86_64.tar.xz"
         sudo cp shellcheck-latest/shellcheck /usr/bin/shellcheck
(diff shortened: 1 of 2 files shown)


manue1 (approved):
LGTM 💯

.github/workflows/main.yml
```diff
@@ -13,14 +13,14 @@ jobs:
     - name: Install Dependencies (Linux)
       run: |
         sudo apt-get update -y
-        sudo apt-get install -y devscripts debhelper moreutils fakeroot jq pigz help2man
+        sudo apt-get install -y devscripts debhelper moreutils fakeroot jq pigz help2man wget
         wget "https://github.com/koalaman/shellcheck/releases/download/latest/shellcheck-latest.linux.x86_64.tar.xz"
```
jsoref, line 17: wouldn't it make more sense to just use `curl`?

jsoref (commented):
closes #923

.github/workflows/main.yml
djdefi: I agree that we should avoid adding another package here

jsoref: Do note that both curl and wget are preinstalled for all supported action runners...

I grew up w/ `wget`, but unless one is doing spidering, `curl` is probably the right choice.

It doesn't really matter to me.

alejndr0: During the testing I could not make curl work due to unexpected behaviour. Later I realised a fake curl was being triggered instead of the actual binary (https://github.com/github/backup-utils/blob/master/test/bin/curl) . As @jsoref mentioned it is not really a new dependency, therefore I would leave it as it is 🙂

bonsohi:
Providing update about this fix - this helped in successfully building `github-backup-utils_3.5.0_all.deb` package which is required as an [asset](https://github.com/github/backup-utils/releases/tag/v3.5.0) for the release! 👏

PR #927, opened 2022-06-15
Only Start Actions if they have been Stopped

Jcambass:
This patch adds a check to guard the starting of actions to only happen if we actually reached the step where we stop actions.


bin/ghe-restore
```diff
@@ -104,7 +104,7 @@ cleanup () {
     update_restore_status "$1"
   fi
 
-  if ghe-ssh "$GHE_HOSTNAME" -- 'ghe-config --true app.actions.enabled'; then
+  if ACTIONS_STOPPED && ghe-ssh "$GHE_HOSTNAME" -- 'ghe-config --true app.actions.enabled'; then
```
Jcambass, line 107: I thought about trying to remove the `ghe-actions-start` step from `cleanup` and add a special `cleanup` method that is only `trap`ed after we've stopped actions but this seemed like the saner solution.

Jcambass, line 107: I thought about trying to remove the `ghe-actions-start` step from `cleanup` and add a special `cleanup_actions` method that is only `trap`ed after we've stopped actions but this seemed like the saner solution.

bin/ghe-restore
```diff
@@ -104,7 +104,7 @@ cleanup () {
     update_restore_status "$1"
   fi
 
-  if ghe-ssh "$GHE_HOSTNAME" -- 'ghe-config --true app.actions.enabled'; then
+  if $ACTIONS_STOPPED && ghe-ssh "$GHE_HOSTNAME" -- 'ghe-config --true app.actions.enabled'; then
```
Jcambass, line 107: I thought about trying to remove the `ghe-actions-start` step from `cleanup` and add a special `cleanup_actions` method that is only `trap`ed after we've stopped actions but this seemed like the saner solution.

boxofyellow (approved):
LGTM: :shipit:

bin/ghe-restore
```diff
@@ -161,6 +161,9 @@ hostname=$(echo "$GHE_HOSTNAME" | cut -f 1 -d :)
 # Show usage with no <host>
 [ -z "$GHE_HOSTNAME" ] && print_usage
 
+# Flag to indicate if this script has stopped Actions.
+ACTIONS_STOPPED=false
```
Jcambass, line 165: @boxofyellow Init this variable here feels a bit arbitrary but I couldn't find a better place. Any idea? 

I also see that we `export` most of the variables but from what I see this is not needed here, should we do it anyway?

If we don't initialise the variable the `if $ACTIONS_STOPPED` will actually evaluate to `true`.

boxofyellow: I don't think we want to `export` it, since we are going we are only interested in using it here within this script.

boxofyellow: Hmmm that being said look what we are doing with `CLUSTER`

https://github.com/github/backup-utils/blob/20db671238089abd45ac32d4ef468639fee273ac/bin/ghe-restore#L198

boxofyellow: But that may be in place to support https://github.com/github/backup-utils/blob/master/share/github-backup-utils/ghe-restore-pages

boxofyellow (approved):
LGTM

bin/ghe-restore
Jcambass: The `export CLUSTER` also confused me but the usage in `ghe-restore-pages` makes it clearer.

PR #939, opened 2022-07-28
Clarify when an application is ready for restore

BlackVegetable:
I'm not overly sold on this location for this information but knowing when the application is sufficiently "set up" for backup restoration to start overwriting things decreases confusion. I'm opening this PR as a discussion point on where to put these clarifications.


docs/usage.md
```diff
@@ -6,7 +6,9 @@ After the initial backup, use the following commands:
    along with full snapshots of all other pertinent data stores.
  - The `ghe-restore` command restores snapshots to the same or separate GitHub
    Enterprise appliance. You must add the backup host's SSH key to the target
-   GitHub Enterprise Server appliance before using this command.
+   GitHub Enterprise Server appliance before using this command. 
+   Your Github appliance is ready to be replaced when you have reached the my.domain.name:8443/setup/settings page. 
+   (This is a very convenient location to use to upload the public SSH key for the backup host.)
```
djdefi, line 11: Not sure I follow this change.  What is being replaced?

Uploading a SSH key for use by the backup utils host is a required step, and must be done from `setup/settings` (unless using the API), so this is more than just convenient, it is part of the restore process.

Maybe it would be better to link out to the GitHu Docs page at https://docs.github.com/enterprise-server/admin/configuration/configuring-your-enterprise/configuring-backups-on-your-appliance and/or https://docs.github.com/enterprise-server/admin/installation/setting-up-a-github-enterprise-server-instance/setting-up-a-staging-instance which discusses the backup and restore process in more detail?

djdefi (changes requested):

PR #940, opened 2022-07-29
Update Dockerfile to buster

djdefi:
Stretch is reaching end of life, so this moves the Dockerfile base ahead to `buster`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM debian:bullseye-slim
+FROM debian:buster-slim
 
 RUN apt-get -q -y update && \
     apt-get install -y --no-install-recommends \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM debian:stretch-slim
+FROM debian:bullseye-slim
```
manue1, line 1: Should we align it with the base OS version of GHES and use `buster` instead?

djdefi: We certainly can do that. I sort of went back and forth on this, and went with the newest. 

On the other hand, we're running tests here in `ubuntu` vs Debian so we are not super tightly coupled to a particular distro/release.

manue1 (approved):

PR #941, opened 2022-07-29
Bump actions versions

djdefi:
Moves to checkout v3 and a newer version of super linter (linting seems to have been broken here for quite some time)

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -7,12 +7,15 @@ on:
     branches: [master]
 
 jobs:
-  build:
+  lint:
     name: Lint Code Base
     runs-on: ubuntu-latest
     steps:
       - name: Checkout Code
         uses: actions/checkout@v3
+        with:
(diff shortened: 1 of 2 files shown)


.github/workflows/lint.yml
```diff
@@ -11,8 +11,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout Code
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
       - name: Lint Code Base
-        uses: docker://github/super-linter:v2.1.1
+        uses: github/super-linter@v4
```
djdefi, line 16: We could alternatively just remove this workflow. It appears to be ignoring Pull Requests targeting master, so it almost never runs.

manue1: Having the shellcheck run might be of some value, so what if we update the trigger? Maybe something like this?
```
on:
  push:
    branches-ignore: [master]
  pull_request:
    branches: [master]
```

manue1 (approved):

PR #946, opened 2022-08-03
Stop restore when a backup without Actions is being restored onto an appliance with Actions enabled

lucavallin:
When a backup without Actions is being restored onto an appliance with Actions enabled, one or more of the secrets/config values used for Actions will be mismatched. Some of the config values I just mentioned are located in places that are not touched by the backup being restored, therefore the (to-be-restored) appliance still has old, pre-restore data laying around (in the database and env variables).

@boxofyellow and I decided to  stop users from restoring backups without Actions onto appliances that do have Actions enabled.

### DIFF
--- a/bin/ghe-restore
+++ b/bin/ghe-restore
@@ -286,7 +286,7 @@ fi
 RELEASE_VERSION=$(ghe-ssh "$GHE_HOSTNAME" -- 'ghe-config --get core.package-version' | cut -d '.' -f 1,2)
 
 # If the backup being restored is from an appliance with Actions disabled, restoring it onto an appliance with Actions enabled will cause
-# mismatches in the secrets needed for Actions which ultimately results in Actions not working properly.
+# mismatches in the secrets needed for Actions which ultimately results in Actions not working properly. Note: xargs is to remove whitespace
 ACTIONS_ENABLED_IN_BACKUP=$(git config -f $GHE_RESTORE_SNAPSHOT_PATH/settings.json --list | grep "app.actions.enabled" | xargs)
 # Get the value of app.actions.enabled
 ACTIONS_ENABLED_IN_BACKUP=${ACTIONS_ENABLED_IN_BACKUP#*=}
--- a/bin/ghe-restore
+++ b/bin/ghe-restore
@@ -287,8 +287,8 @@ RELEASE_VERSION=$(ghe-ssh "$GHE_HOSTNAME" -- 'ghe-config --get core.package-vers
 
 # If the backup being restored is from an appliance with Actions disabled, restoring it onto an appliance with Actions enabled will cause
 # mismatches in the secrets needed for Actions which ultimately results in Actions not working properly. Note: xargs is to remove whitespace
-ACTIONS_ENABLED_IN_BACKUP=$(git config -f $GHE_RESTORE_SNAPSHOT_PATH/settings.json app.actions.enabled | xargs)
-if [[ $ACTIONS_ENABLED_IN_BACKUP != "true" ]] && ghe-ssh "$GHE_HOSTNAME" -- 'ghe-config --true app.actions.enabled'; then
+ACTIONS_ENABLED_IN_BACKUP=$(git config -f $GHE_RESTORE_SNAPSHOT_PATH/settings.json --bool app.actions.enabled | xargs)
+if [[ $ACTIONS_ENABLED_IN_BACKUP != true ]] && ghe-ssh "$GHE_HOSTNAME" -- 'ghe-config --true app.actions.enabled'; then
     echo "Error: Restoring a backup with Actions disabled onto an appliance with Actions enabled is not supported." >&2
     exit 1
 fi
--- a/bin/ghe-restore
+++ b/bin/ghe-restore
@@ -285,17 +285,17 @@ fi
 # Get GHES release version in major.minor format
 RELEASE_VERSION=$(ghe-ssh "$GHE_HOSTNAME" -- 'ghe-config --get core.package-version' | cut -d '.' -f 1,2)
 
-
 # If the backup being restored is from an appliance with Actions disabled, restoring it onto an appliance with Actions enabled will cause
 # mismatches in the secrets needed for Actions which ultimately results in Actions not working properly.
-ACTIONS_ENABLED_IN_BACKUP=$(git config -f $GHE_RESTORE_SNAPSHOT_PATH/settings.json --list | tr . _ | grep "app_actions_enabled" | awk -F"=" '{ print $2 }')
+ACTIONS_ENABLED_IN_BACKUP=$(git config -f $GHE_RESTORE_SNAPSHOT_PATH/settings.json --list | grep "app.actions.enabled" | xargs)
+# Get the value of app.actions.enabled
+ACTIONS_ENABLED_IN_BACKUP=${ACTIONS_ENABLED_IN_BACKUP#*=}
 # If ACTIONS_ENABLED_IN_BACKUP is still empty, then set it to the default value of "false"
-ACTIONS_ENABLED_IN_BACKUP=${ACTIONS_ENABLED_IN_BACKUP:-false}
+ACTIONS_ENABLED_IN_BACKUP=${ACTIONS_ENABLED_IN_BACKUP:-"false"}
 if [[ $ACTIONS_ENABLED_IN_BACKUP == "false" ]] && ghe-ssh "$GHE_HOSTNAME" -- 'ghe-config --true app.actions.enabled'; then
     echo "Error: Restoring a backup with Actions disabled onto an appliance with Actions enabled is not supported." >&2
     exit 1
 fi
-
 
 # Make sure the GitHub appliance has Actions enabled if the snapshot contains Actions data.
 # If above is true, also check if ac is present in appliance then snapshot should also contains ac databases
--- a/bin/ghe-restore
+++ b/bin/ghe-restore
@@ -287,7 +287,7 @@ RELEASE_VERSION=$(ghe-ssh "$GHE_HOSTNAME" -- 'ghe-config --get core.package-vers
 
 # If the backup being restored is from an appliance with Actions disabled, restoring it onto an appliance with Actions enabled will cause
 # mismatches in the secrets needed for Actions which ultimately results in Actions not working properly. Note: xargs is to remove whitespace
-ACTIONS_ENABLED_IN_BACKUP=$(git config -f $GHE_RESTORE_SNAPSHOT_PATH/settings.json --bool app.actions.enabled | xargs || false)
+ACTIONS_ENABLED_IN_BACKUP=$(git config -f $GHE_RESTORE_SNAPSHOT_PATH/settings.json --bool app.actions.enabled | xargs)
 if [[ $ACTIONS_ENABLED_IN_BACKUP != true ]] && ghe-ssh "$GHE_HOSTNAME" -- 'ghe-config --true app.actions.enabled'; then
     echo "Error: Restoring a backup with Actions disabled onto an appliance with Actions enabled is not supported." >&2
     exit 1
--- a/bin/ghe-restore
+++ b/bin/ghe-restore
@@ -289,6 +289,8 @@ RELEASE_VERSION=$(ghe-ssh "$GHE_HOSTNAME" -- 'ghe-config --get core.package-vers
 # If the backup being restored is from an appliance with Actions disabled, restoring it onto an appliance with Actions enabled will cause
 # mismatches in the secrets needed for Actions which ultimately results in Actions not working properly.
 ACTIONS_ENABLED_IN_BACKUP=$(git config -f $GHE_RESTORE_SNAPSHOT_PATH/settings.json --list | tr . _ | grep "app_actions_enabled" | awk -F"=" '{ print $2 }')
+# If ACTIONS_ENABLED_IN_BACKUP is still empty, then set it to the default value of "false"
+ACTIONS_ENABLED_IN_BACKUP=${ACTIONS_ENABLED_IN_BACKUP:-false}
 if [[ $ACTIONS_ENABLED_IN_BACKUP == "false" ]] && ghe-ssh "$GHE_HOSTNAME" -- 'ghe-config --true app.actions.enabled'; then
     echo "Error: Restoring a backup with Actions disabled onto an appliance with Actions enabled is not supported." >&2
     exit 1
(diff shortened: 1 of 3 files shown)


bin/ghe-restore
```diff
@@ -285,6 +285,16 @@ fi
 # Get GHES release version in major.minor format
 RELEASE_VERSION=$(ghe-ssh "$GHE_HOSTNAME" -- 'ghe-config --get core.package-version' | cut -d '.' -f 1,2)
 
+
+# If the backup being restored is from an appliance with Actions disabled, restoring it onto an appliance with Actions enabled will cause
+# mismatches in the secrets needed for Actions which ultimately results in Actions not working properly.
+ACTIONS_ENABLED_IN_BACKUP=$(git config -f $GHE_RESTORE_SNAPSHOT_PATH/settings.json --list | tr . _ | grep "app_actions_enabled" | awk -F"=" '{ print $2 }')
```
lucavallin, line 291: @boxofyellow @conorsloan The `settings.json` file which contains the config I need has a structure like this:

```
[app "actions"]
    enabled = false
```

Which doesn't seem like JSON to me, but more like TOML or ini, unless JSON can in fact be written like this. The `git config` command above parses and lists the settings in the file. The "hierarchy" is then removed and replaced by key/value pairs in which the key has a structure like `<section-name>_<key`> (example: `app_actions_enabled`).

I am not sure this is the best way to read the settings or the best way to parse the file in bash and use the value I need, therefore I kindly request your advice 😄

bin/ghe-restore
```diff
@@ -285,6 +285,16 @@ fi
...
+if [[ $ACTIONS_ENABLED_IN_BACKUP == "false" ]] && ghe-ssh "$GHE_HOSTNAME" -- 'ghe-config --true app.actions.enabled'; then
```
boxofyellow, line 292: Is checking "false" safe here?  What if the value is not in file?

bin/ghe-restore
boxofyellow: That might work Artifact Cache dealt with this problem like this.  Maybe something like that work work here too?

https://github.com/github/backup-utils/blob/11143678e531ba0e88af0e59db4db3b2ab986bf6/bin/ghe-restore#L292

bin/ghe-restore
lucavallin: Good point! I have checked backups from appliances with actions enabled and disabled and this value seems to always be set. But it makes sense to account for the key not being present to be on the safe side.

bin/ghe-restore
conorsloan: The one thing I would recommend here is to use parameter expansion for this, since it splits up the long chain of command and avoids use of `tr` and `awk` in favour of bash features. 

e.g. 
``` bash 
# Explicitly grabbing config in a familiar format with grep (outputs "app.actions.enabled=false")
var=$(git config -f $GHE_RESTORE_SNAPSHOT_PATH/settings.json --list || grep "app.actions.enabled") 
# Replace the .'s with _'s using replace (outputs "app_actions_enabled=false")
var=${var//./_}
# Remove everything before the = (outputs "false")
var=${var#*=}
```

You can use more descriptive names here or a temporary variable. e.g.
* assign repeatedly to ACTIONS_ENABLED_IN_BACKUP
* create a function and use a local variable 
* use a tmp variable to build the string and unset it 

This is personal preference more than a hard rule, but it tends to be easier to grok and debug than long pipelines.

lucavallin: If that's possible, this might be a better solution than 4 different unix commands piped together.  `ghe-mssql-console` however seems to only be available on the appliance, how is `ghe-restore`  going to be able to query the database? For example, if I run `ghe-restore` on my machine, I should be able to just restore the backup I have locally.

Also, this line is running a query on the appliance, but we need to know if the backup we are restoring has Actions enabled or not. The status of actions on the appliance is something we can already retrieve with `ghe-ssh "$GHE_HOSTNAME" -- 'ghe-config --true app.actions.enabled'`, used below.

conorsloan: At the very least I would:
* Grep before tr, since that matches the pattern actually present in the file 
* Consider using the second parameter expansion `${var#*=}` to remove everything before = instead of `awk`

bin/ghe-restore
conorsloan: FYI: Another cool thing you can do here rather than explicitly checking for nil is use the aforementioned parameter expansion to set a default value; there are two ways:

``` bash
var=${var:-false} #explicitly set to itself or default
${var:=false} # If empty, var is assigned to default value
```

That is, if the correct behaviour is empty == false

bin/ghe-restore
tiwarishub: >ghe-mssql-console however seems to only be available on the appliance, how is ghe-restore going to be able to query the database?

Correct. But in artifactCache [here](https://github.com/github/backup-utils/blob/master/bin/ghe-restore#L292). We first logging to ghe instance (see in last ``ghe-ssh "$GHE_HOSTNAME" /bin/bash``) and then running ``ghe-mssql-console``

bin/ghe-restore
lucavallin: Updated!

bin/ghe-restore
lucavallin: @tiwarishub thanks for your comment! This seems to be a valid approach when we want to check if the appliance which we are restoring the backup onto has ArtifactCache enabled. In my case, I want to check if the backup that is being restored has Actions enabled, the other way around therefore. I see that for ArtifactCache we are also checking some `.bak` files, that might work here.

bin/ghe-restore
```diff
@@ -285,6 +285,18 @@ fi
...
 RELEASE_VERSION=$(ghe-ssh "$GHE_HOSTNAME" -- 'ghe-config --get core.package-version' | cut -d '.' -f 1,2)
 
+# If the backup being restored is from an appliance with Actions disabled, restoring it onto an appliance with Actions enabled will cause
+# mismatches in the secrets needed for Actions which ultimately results in Actions not working properly.
+ACTIONS_ENABLED_IN_BACKUP=$(git config -f $GHE_RESTORE_SNAPSHOT_PATH/settings.json --list | grep "app.actions.enabled" | xargs)
+# Get the value of app.actions.enabled
+ACTIONS_ENABLED_IN_BACKUP=${ACTIONS_ENABLED_IN_BACKUP#*=}
+# If ACTIONS_ENABLED_IN_BACKUP is still empty, then set it to the default value of "false"
+ACTIONS_ENABLED_IN_BACKUP=${ACTIONS_ENABLED_IN_BACKUP:-"false"}
+if [[ $ACTIONS_ENABLED_IN_BACKUP == "false" ]] && ghe-ssh "$GHE_HOSTNAME" -- 'ghe-config --true app.actions.enabled'; then
```
lucavallin, line 295: @conorsloan Thanks for the advice on bash, it's looking better now. I don't actually need to replace `.` with `_`, but I had to pipe the result of `grep` to `xargs`, which trims the output (when the keys are not present in `settings.json` grep seems to return whitespaces and carriage returns that break the script unexpectedly.

@boxofyellow  & @tiwarishub: I am now looking into the local backup files to see if we can use a similar approach for this one as we do for ArtifactCache.

conorsloan: Btw sorry I didn't spot this earlier, but can't you just do:

``` bash
ACTIONS_ENABLED_IN_BACKUP=$(git config -f <file> app.actions.enabled)
ACTIONS_ENABLED_IN_BACKUP=${ACTIONS_ENABLED_IN_BACKUP:-false}
```

git config supports grabbing the value of a specific path this way

conorsloan: I tested this locally and it seemed happy enough:

``` bash
conorsloan@Arcanine ~ % cat /tmp/test.txt
[app "actions"]
	enabled = true
	disabled = false
conorsloan@Arcanine ~ % git config -f /tmp/test.txt app.actions.enabled
true
conorsloan@Arcanine ~ % git config -f /tmp/test.txt app.actions.disabled
false
```

lucavallin: Amazing, this works, I wasn't aware of it! I updated it but left the `xargs` for the same reason as earlier (breaks when the value is not in the file)

bin/ghe-restore
```diff
@@ -285,6 +285,16 @@ fi
 # Get GHES release version in major.minor format
 RELEASE_VERSION=$(ghe-ssh "$GHE_HOSTNAME" -- 'ghe-config --get core.package-version' | cut -d '.' -f 1,2)
 
+# If the backup being restored is from an appliance with Actions disabled, restoring it onto an appliance with Actions enabled will cause
+# mismatches in the secrets needed for Actions which ultimately results in Actions not working properly. Note: xargs is to remove whitespace
+ACTIONS_ENABLED_IN_BACKUP=$(git config -f $GHE_RESTORE_SNAPSHOT_PATH/settings.json app.actions.enabled | xargs)
```
conorsloan, line 290: Do you still need `xargs`?

bin/ghe-restore
lucavallin: Update: I wasn't able to find any obvious way to use data/files from the backup's database snapshot to tell if actions is enabled or not. Using the value in `settings.json` seems like the safest option to me at the moment. I will close this conversation but feel free to reopen it as needed!

bin/ghe-restore
```diff
@@ -285,6 +285,16 @@ fi
...
+# If ACTIONS_ENABLED_IN_BACKUP is still empty, then set it to the default value of "false"
+ACTIONS_ENABLED_IN_BACKUP=${ACTIONS_ENABLED_IN_BACKUP:-"false"}
+if [[ $ACTIONS_ENABLED_IN_BACKUP != "true" ]] && ghe-ssh "$GHE_HOSTNAME" -- 'ghe-config --true app.actions.enabled'; then
```
conorsloan, line 293: Nit: If you're going to check `!= "true"` rather than `== "false"`, then you probably don't need to assign "false" as default on line 292 anymore

bin/ghe-restore
conorsloan: Ah I just spotted this comment! Ignore my comment about this, then. 
It surprises me that this breaks though... is there trailing whitespace in the file itself?

bin/ghe-restore
lucavallin: Good one, that was my intention but I did not commit it yet :)

bin/ghe-restore
conorsloan: I see you mentioned this above, so we can discuss it there and you can ignore this comment!

conorsloan: I'm curious about this, if you have a chance could you run without xargs but with the `-x` flag set? 
You can just do `bash -x ghe-restore` or something... you should be able to see what exactly this is set to. 

Alternatively, you could just run this command in your terminal `git config -f $GHE_RESTORE_SNAPSHOT_PATH/settings.json some.missing.key | wc -c` 

Basically, I wouldn't expect a missing value to give you anything but an empty string and I'm wondering if this is a sign of some other bug!

conorsloan: (I'm happy to try this out myself but I don't actually know where `$GHE_RESTORE_SNAPSHOT_PATH` points to so I can't reproduce) 😄

lucavallin: `wc -c` returns 7 spaces and a `0`.  `bash -x ghe-restore` instead returns this instead (including only the last part before my `exit 1` that I put there for testing.

```
++ git config -f /Users/lucavallin/Projects/backup-utils-tainted/data/20220801T095510/settings.json --bool app.actions.enabled
+ ACTIONS_ENABLED_IN_BACKUP=
+ cleanup failed
+ '[' -n failed ']'
+ update_restore_status failed
+ false
+ echo failed
+ ghe-ssh lucavallin-0c579e4255abff420.ghe-test.ninja:122 -- 'sudo sponge '\''/data/user/common/ghe-restore-status'\'' >/dev/null'
+ false
+ true
+ ghe-ssh --clean
```

bin/ghe-restore
```diff
@@ -285,6 +285,14 @@ fi
 # Get GHES release version in major.minor format
 RELEASE_VERSION=$(ghe-ssh "$GHE_HOSTNAME" -- 'ghe-config --get core.package-version' | cut -d '.' -f 1,2)
 
+# If the backup being restored is from an appliance with Actions disabled, restoring it onto an appliance with Actions enabled will cause
+# mismatches in the secrets needed for Actions which ultimately results in Actions not working properly. Note: xargs is to remove whitespace
+ACTIONS_ENABLED_IN_BACKUP=$(git config -f $GHE_RESTORE_SNAPSHOT_PATH/settings.json --bool app.actions.enabled | xargs)
+if [[ $ACTIONS_ENABLED_IN_BACKUP != true ]] && ghe-ssh "$GHE_HOSTNAME" -- 'ghe-config --true app.actions.enabled'; then
```
conorsloan, line 291: No change needed here, but just an FYI: bash doesn't have booleans. You're still assigning "true", "false" or "" to a string variable here and checking it matches the string "true" (which is fine!)

bin/ghe-restore
conorsloan: Very interesting! I think defending against whitespace is probably appropriate in this case, then 👍 . 
For my own curiosity, would you mind sending me a copy of your settings.json file? 😄

lucavallin: Of course, will send it to you via Slack!

bin/ghe-restore
lucavallin: Oh, thanks for yet another bash-tip, I'll keep it in mind! ❤️

lucavallin:
@boxofyellow mentioned that some tests are failing, and actually I haven't added any either. I will fix these problems before we can merge it.

lucavallin:
The tests are currently failing because in CI we create a fake `settings.json` that just contains the message `fake ghe-export-settings data`. `git config` does not appreciate.

test/testlib.sh
```diff
@@ -594,3 +594,13 @@ cleanup_moreutils_parallel() {
     unlink "$ROOTDIR/test/bin/parallel"
   fi
 }
+
+# setup_actions_enabled_in_settings_json writes settings for the Actions app to settings.json
+# it accepts true or false as first argument to enable or disable actions in settings.json
+setup_actions_enabled_settings_for_restore() {
```
boxofyellow, line 600: we use this to read the data 
```
ACTIONS_ENABLED_IN_BACKUP=$(git config -f $GHE_RESTORE_SNAPSHOT_PATH/settings.json --bool app.actions.enabled | xargs || false)
```
should we use `git config` to set the data in the file?

PR #954, opened 2022-09-08
refactor: ghe-cluster-each --print, --ip, --uuid flags

tyliec:
Refactor `ghe-cluster-each` calls to instead use `ghe-cluster-nodes` as appropriate. As part of our 3.7 release, certain flags are being removed. This PR refactors `backup-utils` to use the new method instead.

/cc @donal


donal:
The linter failures are from existing files (and the test isn't required), so we're going to ignore (these failures can be fixed in a separate PR).

donal (approved):

PR #974, opened 2022-11-15
Add jq to requirements

djdefi:
Since `jq` is not always installed by default.

### DIFF
--- a/docs/requirements.md
+++ b/docs/requirements.md
@@ -6,7 +6,7 @@ storage and must have network connectivity with the GitHub Enterprise Server app
 ## Backup host requirements
 
 Backup host software requirements are modest: Linux or other modern Unix operating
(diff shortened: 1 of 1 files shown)


kozakmc:
- [ ]

donal (approved):
This change is :+1:

Although I can only find `jq` being used in one place: https://github.com/github/backup-utils/blob/master/bin/ghe-restore#L542

And I wonder if it wouldn't be better to remove the use of `jq` and therefore remove it as dependency.

(On the other hand, `jq` is very useful and it's more likely that we'll make further use of it.)

djdefi:
>And I wonder if it wouldn't be better to remove the use of jq and therefore remove it as dependency.

Yeah I had that thought too... I'll merge this for now and we can follow up if we want to remove this one usage.

PR #982, opened 2022-12-01
Capture backup host info in backup output

djdefi:
In support, often we do not have immediate visibility into the specifications and OS of the host that backup-utils is running when we backup receive logs or output. 

This PR adds OS, CPU cores, and Memory of the system running backup-utils into the `ghe-backup` output.

I decided to not hide this behind verbose `-v` mode, as I think this information is useful in the default output.

Example:

```
Starting backup of github.example.com with backup-utils v3.7.0 in snapshot 20221201T220633
Running on: Ubuntu 22.04.1 LTS
CPUs: 5
Memory total/used/free+share/buff/cache: 7850/309/5828+322/1713/7023
...
```

cc: @github/ghes-lifecycle @github/support-squad-infrastructure


shcorbett:
thanks @djdefi - this is great. We are also looking at pre-checks around this so this might also tie into that (although I know in this case you are just outputting to log) cc: @gamefiend

Tazminy94 (approved):

Dinkar2014 (approved):

PR #995, opened 2023-01-26
Set a 2 minute timeout on starting cron

djdefi:
If `service cron start` hangs, we don't really want to wait around forever or terminate the entire restore this late in the process.

Adding a 2 minute timeout onto `service cron start` would allow the restore to continue if a service hang occurs, and also log the Warning.

### DIFF
--- a/bin/ghe-restore
+++ b/bin/ghe-restore
@@ -89,11 +89,11 @@ done
 start_cron () {
   echo "Starting cron ..."
   if $CLUSTER; then
-    if ! ghe-ssh "$GHE_HOSTNAME" -- "ghe-cluster-each -- sudo service cron start"; then
+    if ! ghe-ssh "$GHE_HOSTNAME" -- "ghe-cluster-each -- sudo timeout 120s service cron start"; then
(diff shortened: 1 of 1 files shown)


rohthegreat:
Let me check these and I will update you on what is wrong with each one.

rohthegreat:
This is one of the errors 

2023-01-26 23:02:41 [ERROR]   Found errors in [shellcheck] linter!
2023-01-26 23:02:41 [ERROR]   Error code: 1. Command output:
2023-01-26 23:02:45 [ERROR]   ERRORS FOUND in BASH:[1]
2023-01-26 23:02:48 [FATAL]   Exiting with errors found!
I found at least four errors on our side so what do you think would be the fix to these?

rohthegreat (approved):
Ok I approve this change for now but the test file in the commits is in a X and was completed with the exit code 2.

gamefiend (approved):
LGTM!

djdefi:
@rohthegreat all of the linter failures are of `info` or `style` severity, and occur in lines outside of my changes.

These should probably be fixed up at some point, but are not critical for now.

PR #998, opened 2023-01-30
Add zstd to backup-utils

daxamin:
This PR replaces `gzip` with `zstd`.

### Testing

Tested on `gheboot 3.7`
- Populate some data into the appliance using `ghes-populator`
```
./ghes-populator \
create \
--dry-run=false \
--number-of-comments-per-issue=1 \
--number-of-issues-per-repository=5 \
--number-of-repositories-per-user=5 \
--number-of-organizations=1 \
--number-of-sample-repositories=20 \
--number-of-workers-for-populating=5 \
--number-of-users-per-organization=50 \
--number-of-users=50 \
--ghes-url="https://daxamin-0ce2ce8fe5dd4c0ac.qaboot.net" \
--ghes-username="ghe-admin" \
--ghes-password="XXXX" \
--ghes-token="XXXX"
```
- Install `zstd` on gheboot
    - `wget http://ftp.de.debian.org/debian/pool/main/libz/libzstd/zstd_1.3.8+dfsg-3+deb10u2_amd64.deb`
    - `sudo apt install ./zstd_1.3.8+dfsg-3+deb10u2_amd64.deb`
- Take backup `./ghe.backup` after configuring `backup.config`
- Restore using the snapshot `./ghe-restore daxamin-0ce2ce8fe5dd4c0ac.qaboot.net`

### Related items
https://github.com/github/ghes/issues/997

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -13,14 +13,14 @@ jobs:
     - name: Install Dependencies (Linux)
       run: |
         sudo apt-get update -y
-        sudo apt-get install -y devscripts debhelper moreutils fakeroot jq pigz help2man
+        sudo apt-get install -y devscripts debhelper moreutils fakeroot jq pigz help2man zstd
         wget "https://github.com/koalaman/shellcheck/releases/download/latest/shellcheck-latest.linux.x86_64.tar.xz"
         tar --xz -xvf "shellcheck-latest.linux.x86_64.tar.xz"
         sudo cp shellcheck-latest/shellcheck /usr/bin/shellcheck
       if: matrix.os != 'macos-latest'
     - name: Install Dependencies (macOS)
       run: |
-        brew install gnu-tar shellcheck jq pigz coreutils gnu-sed gnu-getopt wget
+        brew install gnu-tar shellcheck jq pigz coreutils gnu-sed gnu-getopt wget zstd
         brew install moreutils gawk
       if: matrix.os == 'macos-latest'
     - name: Get Sources
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,6 +9,7 @@ RUN apt-get -q -y update && \
     git \
     moreutils \
     gawk \
+    zstd \
     && rm -rf /var/lib/apt/lists/*
 
 COPY ./ /backup-utils/
--- a/Dockerfile.alpine
+++ b/Dockerfile.alpine
@@ -9,7 +9,8 @@ RUN apk --update --no-cache add \
     bash \
     gawk \
     procps \
-    coreutils
+    coreutils \ 
+    zstd
 
 COPY ./ /backup-utils/
 WORKDIR /backup-utils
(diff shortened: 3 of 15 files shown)


daxamin:
@gamefiend @elee 👋 

I am not seeing any major improvement in backup and restore times using `zstd` vs `gzip`. Here's some data ->

Data size: 2500 issues, 50 repos, 100 users.
#### `gzip` (master branch)
- backup size: 73.63 GB
- backup time: 806 sec
- restore time: 3990 sec

#### `zstd`
- backup size: 73.64 GB
- backup: 838 sec
- restore: 4923 sec

I am wondering what the next steps are here? 🤔 
- I am not sure why we are seeing a slight increase in the time but I can dig more to find the answer
- Or we can give users the option to use one of `gzip` and `zstd` using some command-line flag (and let `gzip` be the default for now).

Appreciate any thoughts! 🙏

gamefiend:
> I am wondering what the next step here could be? 🤔
> 
> * I am not sure why we are seeing a slight increase in the time but I can dig more to find the answer
> * Or we can give users the option to use one of `gzip` and `zstd` using some command-line flag (and let `gzip` be the default for now).
> 
> Appreciate any thoughts! 🙏

Thanks for running with this @daxamin ! My first guess at why it would be slower is that (from what little I know of the tool) zstd runs most optimally when it has been trained on the files that it is working on. Not being trained on the files it is backing up could explain the difference in performance.  @elee am I understanding this correctly?

I wouldn't say we give users the options to choose between the two unless we are giving them a clear reason to use one over the other. We need to dig into the issue a little further. We might need to look at zstd config/training to get its best performance.

daxamin:
I did some experiments with `pzstd` and it's `-p, --processes   # : number of threads to use for (de)compression (default:<numcpus>)` flag. Result's below ⬇️ 

#### `gzip` (master branch)
- backup size: 4.81 GB
- backup time: 289s
- restore time: 644s

#### `pzstd`
- backup time: 287s
- restore time: 688s

No significant change in time duration if I supply different values for `-p/--processes` (E.g. `-p 8`, `-p 16`, `-p 32`)

elee:
Thanks for trying this out @daxamin -- I think for those not playing along at home a change not in this PR is the flip from `pigz` to `zstd` on a one-off appliance for testing, I believe.

> Not being trained on the files it is backing up could explain the difference in performance. @elee am I understanding this correctly?

I think that's a strong possibility, but we'd likely have to train per compressed file given the nature of mysqldump vs. git vs. RRD are likely to be different.  Maybe a universal dictionary is possible for all appliance files?

> Is it possible that the test dataset just does not have that much data that is actually compressed? It looks like only mysql backups and certain audit log backups would even go through zstd. Do we maybe need to have more data in those places to get a better comparison?

I am wondering too if `73.63 GB` is not representative of enough compressed data in a backup 🤔 

I don't know if there's a documented list of what is compressed in the archive and what is not but we could probably build that relatively easily.  I also wonder if our zstd version (for the backup-utils host in the Dockerfile) is built with `assets` enabled which is shown to be a performance degrader https://github.com/facebook/zstd/pull/3154

shcorbett:
Adding reference to this public repo issue - we may want to relook at this to see if any improvements can be seen now that we have made a lot more modifications to backup/restore --> https://github.com/github/backup-utils/issues/490

PR #1002, opened 2023-02-01
Patch 2 - LINTED-BASH

rohthegreat:
Runs all codes for LINTED-BASH I can assure you that the fix has been made and we need to re-run the jobs at once.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -12,12 +12,12 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout Code
-        uses: actions/checkout@v3
+        uses: actions/checkout@v3.1.2
         with:
           # Full git history is needed to get a proper list of changed files within `super-linter`
           fetch-depth: 0
       - name: Lint Code Base
         uses: github/super-linter@v4
         env:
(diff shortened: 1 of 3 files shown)


rohthegreat:
@djdefi I am awaiting your approval for this change PR

.github/workflows/lint.yml
```diff
@@ -12,12 +12,12 @@ jobs:
...
-        uses: actions/checkout@v3
+        uses: actions/checkout@v3.1.2
         with:
           # Full git history is needed to get a proper list of changed files within `super-linter`
           fetch-depth: 0
       - name: Lint Code Base
         uses: github/super-linter@v4
         env:
-          VALIDATE_ALL_CODEBASE: false
+          VALIDATE_ALL_CODEBASE: True
```
djdefi, line 22: We don't want to make this change, as it would be quite disruptive for most contributions. We only want files involved a PR to be linted. If you would like to lint the codebase in it's entirety I suggest doing that on a local clone with `shellcheck`.

rohthegreat: ok.

rohthegreat:
Closing PR.

PR #1008, opened 2023-02-07
Sync private to public

wirecat:
Hello, this is a result of following our instrunctions on syncing our private repo to the public one during a feature release RC.


ownership.yaml
```diff
@@ -0,0 +1,49 @@
+---
```
djdefi, line 1: We probably want to exclude this file

wirecat: Removed.

gamefiend:
Can we ignore the workflows in `.github` for the time being? Other than that it looks good.

djdefi:
Wouldn't that effectively disable CI here? We probably still need the main and Docker build jobs at the very least.

wirecat:
@djdefi & @gamefiend I removed the new additional workflows, but kept any updates to the existing ones.

gamefiend:
> Wouldn't that effectively disable CI here? We probably still need the main and Docker build jobs at the very least.

Sorry, really what I meant was to not add workflows from private to public that don't exist in public. Clarified this with @wirecat in Slack.

RELEASING.md
```diff
@@ -15,11 +15,24 @@ Only repo administrator is allowed to run the release script, otherwise it will
 Prior to making a release,
 
 1. Sync any changes that have been merged to backup-utils-private into this repository.
-1. Go through the list of open pull requests and merge any that are ready for merging.
-1. Go through the list of closed pull requests since the last release and ensure those that should be included in the release notes:
-  - have a "bug", "enhancement" or "feature" label,
-  - have a title that clearly describes the changes in that pull request. Reword if necessary.
-1. Perform a dry run (add `--dry-run` to one of the commands below) and verify the version strings are going to be changed and verify the release notes.
+
```
gamefiend, line 18: I'm not sure if we want these changes in the public repo.

wirecat: Should this file be removed from the public repo?

gamefiend: I would vote yes, because we don't make releases from public; we sync it and private and then push the changes out from there.

wirecat: I went ahead and removed it for now.

gamefiend (approved):
LGTM, ty @wirecat!

djdefi (approved):

PR #1019, opened 2023-03-08
ghe-backup and ghe-restore shellcheck fixes

krayon:
This is the first of several PRs to harden and improve the backup scripts, initially focusing on shellcheck compliance, in this case 36 conditions were fixed. I have focused on only the main `ghe-backup` and `ghe-restore` scripts in this PR in order to limit the scope, only touching `share/github-backup-utils/ghe-backup-config` where necessary.

The following is a count of the types of fixes:

```
24  SC2086: Double quote to prevent globbing and word splitting.
 4  SC2004: $/${} is unnecessary on arithmetic variables.
 3  SC2223: This default assignment may cause DoS due to globbing. Quote it.
 2  SC2012: Use find instead of ls to better handle non-alphanumeric filenames.
 1  SC2001: See if you can use ${variable//search/replace} instead.
 1  SC2119: Use ghe_remote_version_required "$@" if function's $1 should mean script's $1.
```

### DIFF
--- a/bin/ghe-backup
+++ b/bin/ghe-backup
@@ -79,7 +79,7 @@ if ! output=$(rsync -a src/ dest1 2>&1 && rsync -av src/ --link-dest=../dest1 de
   exit 1
 fi
 
-if [ "$(ls -il dest1/testfile | awk '{ print $1 }')" != "$(ls -il dest2/testfile | awk '{ print $1 }')" ]; then
+if [ "$(stat -c %i dest1/testfile)" != "$(stat -c %i dest2/testfile)" ]; then
   echo "Error: the filesystem containing $GHE_DATA_DIR does not support hard links." 1>&2
   echo "Backup Utilities use hard links to store backup data efficiently." 1>&2
   exit 1
@@ -100,13 +100,13 @@ cleanup () {
     progress=$(cat ../in-progress)
     snapshot=$(echo "$progress" | cut -d ' ' -f 1)
     pid=$(echo "$progress" | cut -d ' ' -f 2)
-    if [ "$snapshot" = "$GHE_SNAPSHOT_TIMESTAMP" ] && [ "$$" = $pid ]; then
+    if [ "$snapshot" = "$GHE_SNAPSHOT_TIMESTAMP" ] && [ "$$" = "$pid" ]; then
       unlink ../in-progress
     fi
   fi
 
   rm -rf "$failures_file"
-  rm -f ${GHE_DATA_DIR}/in-progress-backup
+  rm -f "${GHE_DATA_DIR}/in-progress-backup"
 
   # Cleanup SSH multiplexing
   ghe-ssh --clean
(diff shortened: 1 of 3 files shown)


bin/ghe-backup
```diff
@@ -79,7 +79,7 @@ if ! output=$(rsync -a src/ dest1 2>&1 && rsync -av src/ --link-dest=../dest1 de
   exit 1
 fi
 
-if [ "$(ls -il dest1/testfile | awk '{ print $1 }')" != "$(ls -il dest2/testfile | awk '{ print $1 }')" ]; then
+if [ "$(stat -c %i dest1/testfile)" != "$(stat -c %i dest2/testfile)" ]; then
```
krayon, line 82: `stat -c %i` is much more efficient and simple here. As a side effect we lose the false positive `SC2012` error.

bin/ghe-backup
```diff
@@ -237,7 +237,7 @@ if [ "$GHE_BACKUP_STRATEGY" = "rsync" ]; then
 fi
 
 if [ "$GHE_PARALLEL_ENABLED" = "yes" ]; then
-  $GHE_PARALLEL_COMMAND $GHE_PARALLEL_COMMAND_OPTIONS -- "${commands[@]}"
+  "$GHE_PARALLEL_COMMAND" "${GHE_PARALLEL_COMMAND_OPTIONS[@]}" -- "${commands[@]}"
```
krayon, line 240: The use of in-quote expanded array elements is _much_ safer that a non-quoted string.

bin/ghe-backup
```diff
@@ -266,16 +266,16 @@ if [ -z "$failures" ]; then
...
+echo "Runtime: $((END_TIME - START_TIME)) seconds"
 
 echo "Completed backup of $GHE_HOSTNAME in snapshot $GHE_SNAPSHOT_TIMESTAMP at $(date +"%H:%M:%S")"
 
 # Exit non-zero and list the steps that failed.
 if [ -z "$failures" ]; then
   ghe_remote_logger "Completed backup from $(hostname) / snapshot $GHE_SNAPSHOT_TIMESTAMP successfully."
 else
-  steps="$(echo $failures | sed 's/ /, /g')"
+  steps="${failures// /, }"
```
krayon, line 278: This is much cheaper and easier to read than a `sed`

share/github-backup-utils/ghe-backup-config
```diff
@@ -27,7 +27,7 @@ if [ -n "$GHE_SHOW_VERSION" ]; then
 fi
 
 # Check for "--help|-h" in args or GHE_SHOW_HELP=true and show usage
-# shellcheck disable=SC2120 # the script name is always referenced
+# shellcheck disable=SC2120 # Our arguments are optional and not meant to be the owning script's
```
krayon, line 30: This comment seemed confusing and not relevant to the check it was disabling

gamefiend:
These changes were made on the private repo and will be released into the public repo shortly. Will close this issue to prevent confusion.

PR #1026, opened 2023-04-18
Adding new option to support Pages lazy-build strategy

eptekov:
This PR adds a new option to exclude Pages artifacts from backup and restore:

> If set to 'yes', Pages data will be included in backup and restore. Defaults to 'yes'
> GHE_BACKUP_PAGES=no

Pages Lazy-build was introduced in GHES 3.4 and allows Pages to be rebuild on-demand on a restored host.
According to the ADR, backups should include Pages artifacts by default, with an option to exclude:

> Lazy build is enabled for all enterprise customers and the behavior cannot be disabled. Customers have the choice to include (default) or exclude Pages from their backups.


YiMysty (approved):

Lubot17:
Ok

MC

eptekov:
Closing in favor of https://github.com/github/backup-utils-private/pull/318

PR #1028, opened 2023-05-11
Delete ".DS_Store" file

bonsohi:
### TL;DR
Although this is benign file which somehow got pushed from MacOS - best to get rid of this 👍


daxamin:
@bonsohi Could we add this to `.gitignore` to avoid it being committed/pushed in future? 🙏

bonsohi:
> @bonsohi Could we add this to `.gitignore` to avoid it being committed/pushed in future? 🙏

@daxamin Good suggestion - going to comply! 👏

daxamin (approved):

PR #1047, opened 2023-06-16
Revert 3.6.1 patch

JoeFranks1993:
Revert 3.6.1 patch as this logic is not needed for the 3.6 branch


dooleydevin (approved):

shcorbett (approved):
Approving this to get us back to the state of 3.6 without any changes for encryption key backups as we confirmed with PSE that encryption was not enabled in 3.6. Any other changes that were included as part of this backport should have been reviewed first before being included so we will need to go through that as part of a separate backport and included in next patch

PR #1048, opened 2023-06-16
Revert 3.7.1 patch

JoeFranks1993:
Revert 3.7.1 patch as this logic is not needed for the 3.7 branch


shcorbett (approved):
Approved - reverting the changes for this until we get confirmation that these changes have been tested and no issues. This is still being investigated under this --> https://github.com/github/ghes/issues/6732

PR #1169, opened 2023-12-04
delete /tmp/backup-utils-progress instead of /tmp/backup-utils-progress/*

madbrenner:
# PR Details

## Description

On Thursday 11/30 our nightly backups of our Enterprise GitHub server began failing with the message:
`mkdir: cannot create directory ‘/tmp/backup-utils-progress’: File exists`
After investigating I found our `/tmp` folder had the following files:
```
backup-utils-progress
backup-utils-progress-info
backup-utils-progress-total
backup-utils-progress-type
```
I realized the progress tracking changes merged in https://github.com/github/backup-utils/pull/1164 were not backwards compatible, because the new version expected `/tmp/backup-utils-progress` to be a folder, and when cleaning up old progress would only clean out `/tmp/backup-utils-progress/*`, leaving the FILE `/tmp/backup-utils-progress` on disk, and causing the following `mkdir /tmp/backup-utils-progress` command to fail.
By simply deleting `/tmp/backup-utils-progress` as a whole we can avoid this problem.

### DIFF
--- a/share/github-backup-utils/ghe-backup-config
+++ b/share/github-backup-utils/ghe-backup-config
@@ -663,7 +663,7 @@ prompt_for_confirmation(){
 init-progress() {
 
   if [ -e /tmp/backup-utils-progress ]; then
-    rm -rf /tmp/backup-utils-progress/*
+    rm -rf /tmp/backup-utils-progress
   fi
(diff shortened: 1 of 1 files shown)


jianghao0718:
@madbrenner - thanks for your suggestion. We will adopt this change in our internal repo so you don't have to submit a separate ticket. Thank you again!

